PR #213, opened 2025-08-05
feat(keepsync)!: slim entrypoint w/o wasm, upgrade to automerge v3

jackddouglas:
### Description

Some environments (browser service workers) don't support top-level awaits and `@automerge/automerge` includes a top-level await to initialize the WASM binary. The slim variant externalizes WASM initialization but until now, there was no clean way to initialize WASM externally when using automerge inside `@automerge/automerge-repo`. By upgrading to automerge v3, automerge-repo now includes an interface for initializing WASM.

Keepsync now has two entrypoints: a default entrypoint that auto-initializes WASM, and a `slim` entrypoint (accessed by `@tonk/keepsync/slim`) that exposes `initializeWasm` for use in previously unsupported environments.

TODO:
- [ ] Revisit lazy initialization of WASM for the default entrypoint (it's EXTRMELY hacky right now)

### Other changes

Styling, test fixes, updated keepsync use in some places

### Tested

Yes

### Related issues

- closes TON-1338, TON-1339, TON-1340

### Backwards compatibility

No, BREAKING CHANGES. `configureSyncEngine` is now async

### Documentation

None


packages/keepsync/src/utils/wasmState.ts
```diff
@@ -0,0 +1,14 @@
+import { initializeWasm } from '@automerge/automerge-repo/slim';
+
+let isSlim = false;
+
+export async function waitForWasm(): Promise<void> {
+  if (isSlim) return;
+  await initializeWasm(
+    'https://esm.sh/@automerge/automerge@3.1.1/dist/automerge.wasm'
```
goblinoats, line 8: How would we initializeWasm in a setting where there is no network connection?

goblinoats (approved):

PR #360, opened 2026-01-07
feat(access): tonk access service

jackddouglas:
### Description

- add new `tonk-access-service` crate for UCAN-authorized access to R2 storage
- service is a cloudflare worker written in rust
- ~~DID and keypair assigned at first deployment, stored in key-value store~~
- implemented verification of UCAN invocations
- ~~re-implement AWS V4 signing algo for pre-signed URL generation (based on [https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html))~~
- uses `dialog-s3-credentials` for pre-signed URL generation
- E2E tests, error handling, edge cases, and failure modes

Follows router pattern:
- `GET /{space-did}/index/{blake3hash}  → 307 redirect to presigned GET URL`
- `PUT /{space-did}/index/{blake3hash}  → 307 redirect to presigned PUT URL`
- UCAN in `Authorization` header

### Other changes

None

### Tested

Yes, run `cargo run --package tonk-access-service --bin test-client --features test-client -- --help` for a list of test commands.

E2E test output:
```
     Running `target/debug/test-client e2e-test --service-url 'https://tonk-access-service.tonk.workers.dev'`
=== End-to-End Test ===

Step 1: Getting service info...
  Service: tonk-access-service
  Version: 0.1.0

Step 2: Generating keypairs...
  Space DID: did:key:z6MkhxfZr5zoat6ZCyqS73EdNMs79d6dnMPqagkjwKa6mWYr
  Operator DID: did:key:z6Mkh2LfiNihRzFNxnNEkv9kZtwihPTTfj21hDdX8Gpy2XPd

Step 3: Creating delegation (space -> operator)...
  Delegation CID: bafyreidrcqt3vtm27uxm3wm3gpl36zrqyshgree7lvivuaas5l5bijvhwe

Step 4: Preparing test content...
  Content: "Hello, UCAN!"
  Size: 12 bytes
  Blake3 (addressing): 8373bcf6d184127335d523239505261ece7ff17997afb7e8b5e284c29178e796
  SHA256 (integrity):  yAQ089f4RwTT2t83QiN01CV0BbCTvEv9VJ4+0v13LQE=

Step 5: Sending PUT request...
  Got presigned URL

Step 6: Uploading blob to presigned URL...
  Upload: SUCCESS

Step 7: Sending GET request...
  Got presigned URL

Step 8: Downloading and verifying content...
  Content matches: SUCCESS

=== All Tests Passed! ===

Test artifacts:
  Space Key: KWpjdV0W9j7wrlBZFj28dichLmoRizzyMxdQ3qt2bmk=
  Operator Key: EfObbbFzmY5kieLEJj4IaWz/FLYlPmjkmRDnE+g9kyA=
  Delegation: glhAVeksh4Fw4XZVjLr636DpxZvg03sJ0FJ9cDte0CF5h7ZSw/hEREURXIrxhJ887jeEjrlE8zKpk/3CVkAorxBDA6JhaEg0Ae0B7QETcXN1Y2FuL2RsZ0AxLjAuMC1yYy4xqWNhdWR4OGRpZDprZXk6ejZNa2gyTGZpTmloUnpGTnhuTkVrdjlrWnR3aWhQVFRmajIxaERkWDhH
cHkyWFBkY2NtZGUvaHR0cGNleHD2Y2lzc3g4ZGlkOmtleTp6Nk1raHhmWnI1em9hdDZaQ3lxUzczRWROTXM3OWQ2ZG5NUHFhZ2tqd0thNm1XWXJjbmJm9mNwb2yAY3N1Yng4ZGlkOmtleTp6Nk1raHhmWnI1em9hdDZaQ3lxUzczRWROTXM3OWQ2ZG5NUHFhZ2tqd0thNm1XWXJkbWV0YaBlbm9uY2
VQSci/SmMfKW64wcz3suvP5g==
```

Edge case test output

```
     Running `target/debug/test-client test-all --service-url 'https://tonk-access-service.tonk.workers.dev' --verbose`
=== Running All Test Suites ===

--- Auth Tests ---
=== Authentication & Authorization Tests ===

  [PASS] tampered_signature
  [PASS] expect_audience_mismatch
  [PASS] expect_invocation_expired
  [PASS] expect_proof_not_found
  [PASS] expect_command_mismatch

Auth Tests: PASSED (5 passed, 0 failed)

--- Validation Tests ---
=== Input Validation Tests ===

  [PASS] invalid_base64_auth
  [PASS] invalid_cbor
  [PASS] missing_auth_header
  [PASS] invalid_base64_proofs

Validation Tests: PASSED (4 passed, 0 failed)

--- Delegation Tests ---
=== Delegation Chain Tests ===

  [PASS] direct_delegation
  [PASS] expect_command_mismatch
  [PASS] expect_proof_expired
  [PASS] expect_proof_not_yet_valid
  [PASS] multi_level_delegation

Delegation Tests: PASSED (5 passed, 0 failed)

========================================
          OVERALL TEST SUMMARY          
========================================

  [PASS] Auth Tests
  [PASS] Validation Tests
  [PASS] Delegation Tests

  Total: 3/3 suites passed

  ALL TESTS PASSED!
```

### Related issues

- closes TON-1748, TON-1753, TON-1749, TON-1750, TON-1751, TON-1756, TON-1752, TON-1747, TON-1767, TON-1768, TON-1769, TON-1770, TON-1771, TON-1772

### Backwards compatibility

Yes, changes all in new crate

### Documentation

None

---

## PR-Codex overview
This PR introduces the `tonk-access-service`, a UCAN-authorized service for R2 storage, implementing HTTP handlers for health checks, service info, and blob operations. It also integrates a test client for end-to-end testing.

### Detailed summary
- Added `tonk-access-service` module with HTTP request handlers.
- Implemented `health`, `info`, and `blob` handlers.
- Introduced UCAN verification in the `ucan` module.
- Created R2 operations for generating pre-signed URLs.
- Added a CLI test client for generating keys and testing the service.
- Updated `Cargo.toml` to include new dependencies and features.
- Enhanced error handling with structured error responses.
- Created GitHub workflows for deployment and testing automation.

> The following files were skipped due to too many changes: `rust/tonk-access-service/src/bin/test_client.rs`, `Cargo.lock`


rust/tonk-access-service/src/handlers/info.rs
```diff
@@ -0,0 +1,23 @@
+//! Service info endpoint.
+
+use crate::identity::ServiceIdentity;
+use worker::*;
+
+pub async fn handle(_req: Request, ctx: RouteContext<()>) -> Result<Response> {
+    // Get KV binding
+    let kv = ctx.kv("CONFIG")?;
```
Gozala: What do we need kv for I'm not sure I fully understand it.

rust/tonk-access-service/src/identity.rs
```diff
@@ -0,0 +1,79 @@
...
+pub struct ServiceIdentity {
+    signer: Ed25519Signer,
+}
+
+impl ServiceIdentity {
+    /// Get or create the service identity from KV storage.
+    ///
+    /// On first call, generates a new keypair and stores it.
+    /// On subsequent calls, loads the existing keypair.
+    pub async fn get_or_create(kv: &KvStore) -> Result<Self, IdentityError> {
```
Gozala: Not sure why we need keypair, it feel unnecessary. Either way we probably don't want to generate and store them in KV if do need keypair we should be passing it into as secrets so we can rotate keys as needed.

rust/tonk-access-service/src/r2/presign.rs
Gozala, line 1: I suggest we refactor [access](https://github.com/dialog-db/dialog-db/blob/main/rust/dialog-storage/src/storage/backend/s3/access.rs) module from dialog-backend maybe into separate crate and simply use it instead so we don't have to maintain multiple impls of the same thing.

You could probably start with just importing it from `dialog_backend`, you should be able to create credentials like shown here

https://github.com/dialog-db/dialog-db/blob/main/rust/dialog-storage/src/storage/backend/s3/access.rs#L400-L403

And then create authorizations (which are presigned URL + headers) as follows

https://github.com/dialog-db/dialog-db/blob/main/rust/dialog-storage/src/storage/backend/s3/access.rs#L590-L592

I imagine service would simply serialize Authorization and respond with it and client will then just deserialize it and have a valid authorization.

rust/tonk-access-service/src/ucan/capability.rs
```diff
@@ -0,0 +1,124 @@
...
+            "blob/allocate" => {
+                let space = get_string_arg(arguments, "space")?;
+                let blob = get_blob_arg(arguments)?;
+                Ok(Capability::BlobAllocate(BlobAllocate { space, blob }))
+            }
+            "blob/get" => {
+                let space = get_string_arg(arguments, "space")?;
+                let digest = get_digest_arg(arguments)?;
+                Ok(Capability::BlobGet(BlobGet { space, digest }))
+            }
```
Gozala: I realize I have mislead you as I've failed to consider the fact that in case of dialog blobs are addressed by blake3 hash where's w3up used sha256. Implication is that when dialog attempts to read blob it will want to do it by blake3 hash not by sha256, later is only used for integrity checks.

That is to say I think blob/allocate as per w3up spec is not going to work for us as is and we'd need to also provide desired key we'd want to look it up by (which would be blake3hash). Also note that we'd still want to provide sha-256 digest to ensure that bytes pass integrity check by S3.

Alternative would be to switch dialog to sha-256, but there are various tradedoffs that would need to be considered

cc @cdata

rust/tonk-access-service/src/ucan/verify.rs
```diff
@@ -0,0 +1,218 @@
...
+    cbor_bytes: &[u8],
+    proof_bytes: &[Vec<u8>],
+    service_did: &Ed25519Did,
+) -> Result<VerifiedInvocation, VerificationError> {
+    // Step 1: Parse the invocation
+    let invocation: Invocation<Ed25519Did> = serde_ipld_dagcbor::from_slice(cbor_bytes)
+        .map_err(|e| VerificationError::ParseError(e.to_string()))?;
+
+    // Step 2: Check audience matches our service DID
+    if invocation.audience() != service_did {
```
Gozala: I really should have refreshed my memory on blob/allocate before I suggest it. I apologize for overlooking details that I'm just pointing out now. With that said I think there are two major nuances to consider here and whether what w3up did is worth replicating.

In w3up audience was service provider DID, because consumer obtained delegation from the provider DID when account was registered. So basically once I registered an account & setup a billing I would obtain delegation from the service provider DID to my account DID for various protocols.

That delegation is then used by consumer application to /blob/allocate and bunch of other things. This is why audience of invocation was the provider DID as invocation needs to form a loop.

What I imagined doing in Tonk was that space did itself could form root issuer and there for loop would have terminated with the space DID that is audience of the /blob/allocate would be space DID itself. In return service could provide pre-signed URL.

This would have simplified service as it would not need keypair of it's own all that would be needed is to verify the invocation and if so issue presigned URL scoped by the did of the space.

However that also implies that we would not be able to tell whether there is a payed account associated with a space to cover the cost of hosting, not without making service stateful anyway.

All this is to say that we have two choices:

1. We create some for of registration which creates delegation that can be used to `/blob/allocate`.
2. We pun on the registration part and simply get invocation loops that is in root ucan issuer == subject and in terminal ucan  audience == subject.

Later would imply that we initially allow anyone to write in any space and later start performing checks to ensure only payed customers can do it. I'm inclined to suggest we go with this to keep things simple and introduce complication when we start account registration etc..

P.S. I also think even in cases where we do account registration we probably could do better than making service DID a terminal audience, namely we could leverage something like [ucan/attest](https://github.com/storacha/specs/blob/main/w3-ucan.md) capability that consumer could optionally provide with such invocations providing evidence that space is registered removing need for statefulness when provided, if omitted we could obtain such evidence on demand

rust/tonk-access-service/src/ucan/capability.rs
```diff
@@ -0,0 +1,124 @@
...
+pub struct BlobAllocate {
+    /// Space DID where the blob will be stored
+    pub space: String,
+    /// Blob metadata
+    pub blob: BlobMeta,
+}
+
+/// Arguments for blob/get capability
+#[derive(Debug, Clone)]
+pub struct BlobGet {
```
Gozala: I find this naming pretty confusing, I would expect response to give me back blob while instead it creates authorization that can be used to get blob

Gozala (commented):
I provided more feedback inline, but general sentiment is I think we can simplify design if we make root issuer and terminal audience be space DID instead, this way service need no DID or keypair to operate.

I was also thinking it would be very nice if service acted as a router that is, it receives GET request on `/did:key:zSpace/index/blake3hash` with UCAN in the authorization header and either returns 307 response with redirect to pre-signed URL or 401 if authorization failed. We could do same thing for PUT requests.

With the above I'd probably change commands to `/http/put` and `/http/get` instead and would move paths like `/index/...` or `/memory/...` under UCAN policy.

Once again I apologize for misdirection with `/blob/allocate` spec without taking time to consider this more carefully.

rust/tonk-access-service/src/ucan/verify.rs
jackddouglas: ok we'll go with simple invocation loops for now.

in the future we can have a billing service that issues `ucan/attest` for paid spaces, have the access service check for attestation, and if missing, query billing service on demand so access service stays stateless.

flake.nix
```diff
@@ -259,6 +266,15 @@
...
           };
+
+          # Cargo tests
+          tests = craneLib.cargoTest (
+            commonArgs
+            // {
+              inherit cargoArtifacts;
+              cargoTestExtraArgs = "-- --test-threads=1";
+            }
+          );
```
Gozala, line 277: Do we need this ? Could you add code comment above so reason is captured

```suggestion
          # Cargo tests, need to to remove concurrency because ...
          tests = craneLib.cargoTest (
            commonArgs
            // {
              inherit cargoArtifacts;
              cargoTestExtraArgs = "-- --test-threads=1";
            }
          );
```

result
Gozala, line 1: Is this meant to be here, I'm confused on why if so

rust/tonk-access-service/src/bin/test_client.rs
```diff
@@ -0,0 +1,1604 @@
...
+        // Client that doesn't follow redirects (so we can test 307)
+        let client = Client::builder().redirect(Policy::none()).build()?;
+
+        // Verify service is reachable
+        let _info: ServiceInfo = client.get(service_url).send().await?.json().await?;
+
+        // Generate test keypairs
+        let mut space_seed = [0u8; 32];
+        getrandom::getrandom(&mut space_seed)?;
+        let space_signer = Ed25519Signer::new(SigningKey::from_bytes(&space_seed));
```
Gozala, line 303: Nit: I think you can simply use `SigningKey::generate` instead

rust/tonk-access-service/src/bin/test_client.rs
```diff
@@ -0,0 +1,1604 @@
...
+        );
+
+        // Only add X-UCAN-Proofs header if there are proofs
+        if !proof_bytes.is_empty() {
+            let proofs_header = proof_bytes
+                .iter()
+                .map(|p| BASE64.encode(p))
+                .collect::<Vec<_>>()
+                .join(",");
+            request = request.header("X-UCAN-Proofs", proofs_header);
```
Gozala, line 399: Is this custom thing ? I think we should just be encoding the whole thing as [UCAN Container](https://github.com/ucan-wg/container) and passing it in a header. Maybe those could be ordered as invocation followed by proof chain making it obvious where to start.

rust/tonk-access-service/src/bin/test_client.rs
Gozala, line 1: Nit: probably could be cleaned up from all the print output and such.

rust/tonk-access-service/src/error.rs
```diff
@@ -0,0 +1,158 @@
...
+use worker::Response;
+
+/// Error codes returned by the API.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize)]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+pub enum ErrorCode {
+    // 400 Bad Request - Input validation errors
+    InvalidBase64,
+    InvalidCbor,
+    InvalidArgument,
```
Gozala, line 13: Nit: This is one isn't obvious maybe rename or add code comment

rust/tonk-access-service/src/error.rs
```diff
@@ -0,0 +1,158 @@
...
+    AudienceMismatch,
+    InvocationExpired,
+    ProofNotFound,
+    ProofExpired,
+    ProofNotYetValid,
+
+    // 403 Forbidden - Authorization errors
+    ChainInvalid,
+    CommandMismatch,
+    SubjectNotAllowed,
```
Gozala, line 26: This is confusing, maybe `SubjectMismatch`, assuming it happens when subject is not leading component of the resource path.

rust/tonk-access-service/src/handlers/blob.rs
```diff
@@ -0,0 +1,289 @@
+//! Blob GET/PUT handlers with 307 redirects.
+//!
+//! These handlers implement the router pattern:
+//! - GET /{space_did}/index/{digest} → 307 redirect to presigned GET URL
+//! - PUT /{space_did}/index/{digest} → 307 redirect to presigned PUT URL
+//!
+//! Authorization is provided via headers:
+//! - Authorization: Bearer <base64-dag-cbor-ucan>
+//! - X-UCAN-Proofs: <base64>,<base64>,... (comma-separated delegation proofs)
```
Gozala, line 9: I think it would be simpler and better to just pass Authorization header and encode it all as [UCAN container](https://github.com/ucan-wg/container) ordered as `[invocation, ...proofs]` so it matches the chain order in the invocation. I think it also could be compressed better.

rust/tonk-access-service/src/handlers/blob.rs
```diff
@@ -0,0 +1,289 @@
...
+//!
+//! For PUT requests, optional integrity verification:
+//! - X-Checksum-SHA256: <base64-sha256> - If provided, R2 will verify upload integrity
```
Gozala, line 12: 🤔 I wonder if we could also add `x-amz-checksum-blake3` and assume that hosts that do not support it will ignore it where's hosts that do will perform verification. Although there is a risk that host that does not support algorithm may reject it.

Gozala: Also I believe header is `x-amz-checksum-sha256` not `x-checksum-sha256`, I don't think server will complain but it will also not reject invalid write.

rust/tonk-access-service/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+    let router = Router::new();
+
+    router
+        // Service info endpoint
+        .get_async("/", handlers::info::handle)
+        // Health check
+        .get_async("/health", handlers::health::handle)
+        // Blob routes with 307 redirects
+        .get_async("/:space_did/index/:digest", handlers::blob::handle_get)
+        .put_async("/:space_did/index/:digest", handlers::blob::handle_put)
```
Gozala, line 27: I think we should be more flexible in our routing as in not limit to `/space/index/digest` structure. Reason is we also write things like `/:space_did/memory/local/:branch`  to track latest revision on that branch. I also expect we will end up doing other paths like /:space_did/blob/:digest` for binary blobs. 

Overall I think access service should be flexible and not care about those details, I think invariant it wants to uphold is

```rs
invocation.command == format!("/http/{request.method}") &&
(url.path.starts_with(format!("{invocation.subject}{invocation.args.key}/")) ||
url.path == format!("{invocation.subject}{invocation.args.key}"))
```

That way service just upholds access to whatever UCAN gives access to without having to know or care about details.

we could probably prefix URLs with `did:` or alternatively we could add namespace like `/ucan/` or something like that.

Gozala: Oh and I think we also may need support for `POST` requests (used for listing entries).

rust/tonk-core/tests/sync.rs
Gozala, line 1: Can we please leave this out from this pr, seems unrelated

Gozala (commented):
Apparently I have not completed review. I think some comments may be outdated by now. I'll publish and review current state now

Cargo.toml
```diff
@@ -26,10 +26,11 @@ dialog-storage = { git = "https://github.com/dialog-db/dialog-db.git", branch =
 dialog-query = { git = "https://github.com/dialog-db/dialog-db.git", branch = "tonk-ecs" }
 dialog-artifacts = { git = "https://github.com/dialog-db/dialog-db.git", branch = "tonk-ecs" }
 dialog-common = { git = "https://github.com/dialog-db/dialog-db.git", branch = "tonk-ecs" }
+dialog-s3-credentials = { git = "https://github.com/dialog-db/dialog-db.git", branch = "feat/s3-presign-crate" }
 
 # UCAN
-ucan = { git = "https://github.com/tonk-labs/rs-ucan.git", branch = "fix/wasm-compile", package = "ucan" }
-varsig = { git = "https://github.com/tonk-labs/rs-ucan.git", branch = "fix/wasm-compile", package = "varsig" }
+ucan = { git = "https://github.com/tonk-labs/rs-ucan.git", branch = "jackddouglas/feat/check", package = "ucan" }
+varsig = { git = "https://github.com/tonk-labs/rs-ucan.git", branch = "jackddouglas/feat/check", package = "varsig" }
```
Gozala, line 33: Sidenote: Are there changes here that we need to uplift into mainline rs-ucan ? I think we should try to get them in to avoid having to maintain fork

rust/tonk-access-service/Cargo.toml
```diff
@@ -0,0 +1,62 @@
...
+test-client = ["dep:reqwest", "dep:clap", "dep:tokio", "dep:anyhow", "dep:sha2", "dep:blake3"]
+
+[dependencies]
+# Cloudflare Workers
+worker = "0.6"
+worker-macros = "0.6"
+
+# UCAN (your fork with invocation exports)
+ucan = { git = "https://github.com/tonk-labs/rs-ucan.git", branch = "jackddouglas/feat/check" }
+varsig = { git = "https://github.com/tonk-labs/rs-ucan.git", branch = "jackddouglas/feat/check" }
```
Gozala, line 28: Would it make sense to have workspace level dependency instead to avoid getting out of sync ? Or if this is intentionally not from workspace maybe add a comment about capturing rationale.

rust/tonk-access-service/src/handlers/blob.rs
Gozala: According to claude `aws` will not care about `x-amz-checksum-blake3` header in which case it seems like a good idea to include.

rust/tonk-access-service/src/handlers/blob.rs
```diff
@@ -0,0 +1,289 @@
...
+        ));
+    }
+
+    // 4. Validate space_did matches verified subject
+    if space_did != &verified.subject {
+        return Err(ServiceError::invalid_argument(format!(
+            "URL space ({}) does not match UCAN subject ({})",
+            space_did, verified.subject
+        )));
+    }
```
Gozala, line 119: Seems like there is some there is an opportunity to reduce some redundancy across routes they all basically verify same thing besides request method

Gozala: That is I think `req` method can be used to derive both expected `command` and `Method` that is passed into `presign::presign_url`

rust/tonk-access-service/src/handlers/blob.rs
```diff
@@ -0,0 +1,289 @@
...
+}
+
+/// Parse X-Checksum-SHA256 header if present.
+///
+/// The header value must be base64-encoded SHA256 (32 bytes).
+/// Returns None if header is not present, error if malformed.
+fn parse_checksum_header(req: &Request) -> std::result::Result<Option<Checksum>, ServiceError> {
+    let header = req
+        .headers()
+        .get("X-Checksum-SHA256")
```
Gozala, line 141: Header is `x-amz-checksum-sha256` I don't know what `x-checksum-sha256` will do, but I suspect it will just be ignored.

rust/tonk-access-service/src/handlers/blob.rs
Gozala: I think we probably do not need to require this header at all. Instead it probably should be something optional in the UCAN itself. I'm bit lost in what actual UCANs look like so I suggest following

```json
{
   "iss": "did:key:whoever",
   "sub": "did:key:space",
   "cmd": "/http/get",
   "args": {
       "key": "index/<blake3hash>",
    }
}
```

```json
{
   "iss": "did:key:whoever",
   "sub": "did:key:space",
   "cmd": "/http/put",
   "args": {
       "key": "index/<blake3hash>",
       "checksum": {
           "sha256": { "/": { "bytes": "<sha256_hash>" } },
           // could also have
           "blake3": { "/": { "bytes": "<blake3_hash>" } },
       },
    }
}
```

That way pre-signer adds `x-amz-checksum-` into required headers and it's up to client to ensure those headers are added.

We could also probably have POST endpoint that instead of doing redirects responds with pre-signed URL + headers. I assume that POST endpoint could live `/` or perhaps `/api/ucan`.

Gozala: 🤔 On the other hand it is also compelling to make service less generic and instead have something like

```ts
type Put {
   cmd: "/store/put",
   args: {
      store: String,
      key: Blake3Hash,
      checksum?: Record<string, ChecksumDigest>
   }
}

type Get {
   cmd: "/store/get",
   args: {
      store: String,
      key: Blake3Hash,
      checksum?: Record<string, ChecksumDigest>
   }
}

type Publish {
   "cmd": "/memory/publish",
   "args": {
      memory: String,
      cell: String,
      edition: Uint8Array | null, // opaque edition id like Etag or value Digest, null implies now prior record
      checksum?: Record<string, ChecksumDigest>
   }
}

type Resolve {
   "cmd": "/memory/resolve",
   "args": {
       memory: String,
       cell: String,
    }
} 
```

This way service has a lot more control of how storage is organized and is more aware of storage semantics.

rust/tonk-access-service/src/handlers/blob.rs
```diff
@@ -0,0 +1,289 @@
...
+
+use crate::{
+    error::{ErrorCode, ServiceError},
+    r2::{Checksum, Method, R2Config, presign},
+    ucan::{VerificationError, verify_invocation},
+};
+use base64::Engine;
+use worker::*;
+
+/// GET /{space_did}/index/{digest} → 307 redirect to presigned GET URL
```
Gozala, line 22:
```suggestion
/// GET /{space_did}/{store}/{digest} → 307 redirect to presigned GET URL
```

I think it's more accurate to think about routes like the above as opposed to "index" being fixed. Alternatively it may be even better to structure as `/:space/store/:store_name/:digest`

rust/tonk-core/examples/node/integration/websocket.test.js
Gozala, line 1: Seems like this should not be part of this PR

rust/tonk-core/examples/shared/test-server.js
Gozala, line 1: Seems like this should not be part of this PR

rust/tonk-core/tests/node-sync/run-sync-test.js
Gozala, line 1: Seems like this should not be part of this PR

Gozala (approved):
Approving with couple of caveats I'd like to get addressed before this merges

1. I believe `x-checksum-sha256` is not a supported header, and seems like we should not ever require it and instead pass checksum as ucan arg.
2. Lets not include changes to tonk-core as those seem unrelated and better land them independently.
3. Can you please add readme.md that describes protocol in terms routes and expected Request / Response, including what is the UCAN invocation format on each one.

Everything else could probably done as followups as needed. Please do note however that current implementation will not work for transactional memory

PR #364, opened 2026-01-14
chore: Move all Rust builds to crane

cdata:
This change re-writes our Flake builds of Rust code to use [Crane](https://crane.dev). Crane is a helper library for building Rust projects that makes it possible to build Cargo dependencies as a separate derivation. This is a nuanced but important distinction from the built-in Rust crate builder, and the upshot is that the dependencies only need to be rebuilt when `Cargo.lock` changes (without Crane, technically the dependencies could be re-used except that their derivation hash is derived from the hash of the dependent crate where the actual changes are happening so they change each time).

Since using Crane correctly involves some nuance, I factored its use out into a dedicated module. The module exports helpers:

- `buildCrate` for building Rust crates in the general case
- `buildTrunkCrate` for building Trunk-specific crates e.g., `tonk-ui`
- `cargoChecks` which includes `cargo-clippy` and `rustfmt` checks
- `buildTestArchive` builds a `cargo-nextest` archive for a given test configuration

Also in this change: I cleaned up some of the `menu.nix` stuff and added a helper for authoring menu commands intended to run tests.

All told, the `flake.nix` is ~150 lines shorter after this change (and hopefully surfaces all the most interesting bits). And perhaps more importantly: the current `test.yml` workflow is almost 15 minutes faster!

---

## PR-Codex overview
This PR focuses on updating dependencies, enhancing the build process, and refining configuration files for the `tonk` project. It includes improvements to the Nix setup and adjustments to the `Cargo.toml` dependencies.

### Detailed summary
- Updated `wasm-bindgen` and `worker` dependencies to newer versions.
- Changed the `main` path in `wrangler.toml`.
- Enhanced Nix build commands with `--accept-flake-config`.
- Modified GitHub Actions workflows for deployment processes.
- Added new helper functions in `menu.nix` for shell menu management.
- Introduced a new Nix overlay for `esbuild`.
- Updated `Cargo.lock` to reflect new dependency versions.
- Removed obsolete comments and cleaned up configuration files.


nix/menu.nix
```diff
@@ -104,7 +106,39 @@ let
...
+    nix build .#${package}
+
+    TESTS_PATH=$(nix eval .#${package}.outPath --raw)
+
+    cargo nextest run \
+      --workspace-remap ./ \
+      --archive-file "$TESTS_PATH/${package}.tar.zst" \
+  '';
+
+  menuTestEnv = with pkgs; lib.optionals stdenv.isLinux {
```
Gozala, line 131: What happens if it's not linux ? Should there be fallback to use whatever's in the system ?

Gozala (approved):

nix/menu.nix
cdata: If it's not Linux, presumably you have your own `chromedriver` + compatible Chrome installed and we must pray that your environment is set up to make it work. That's the price of mutability!

PR #370, opened 2026-01-20
chore: Build all CF assets with Nix and deploy pages

cdata:
This change consolidates the `tonk-access-service` and `tonk-ui` under a single Nix package which can subsequently be deployed to Cloudflare in one go.

It was a bit tricky to get this to work as intended. `worker-build` has a few side-effects: if any of `wasm-bindgen-cli`, `wasm-opt` or `esbuild` at specific version ranges are not found, `worker-build` will attempt to download them. This is not okay with Nix, which is running builds in a sandbox. It's also really confusing DX outside of Nix IMO, but I digress.

I moved the `wrangler.toml` to the root of the project because it now depends on artifacts that come from multiple crates in order for a deploy to work. In a local development environment, you can drop into a dev shell (`nix develop`) and then run `wrangler build` from the root of the project to build the whole Cloudflare deployment as a dry run (`wrangler` will use `nix` under the hood).

---

## PR-Codex overview
This PR focuses on upgrading dependencies, improving build configurations, and refining the deployment process for the `tonk-access-service` and associated components.

### Detailed summary
- Updated `worker` and `worker-macros` versions to `0.7.4`.
- Updated `wasm-bindgen` to version `0.2.108`.
- Changed the build command in `wrangler.toml` for asset management.
- Modified CI configuration in `.github/workflows` to accept flake configurations.
- Adjusted paths for asset directories in `wrangler.toml`.
- Enhanced `nix` configurations to manage `esbuild` and `wasm-bindgen-cli`.
- Updated various dependencies in `Cargo.lock` for compatibility and improvements.


nix/rust.nix
```diff
@@ -36,6 +36,22 @@ let
...
+      pname = "wasm-bindgen-cli";
+      version = "0.2.108";
+      # hash = "sha256-M6WuGl7EruNopHZbqBpucu4RWz44/MSdv6f0zkYw+44=";
+      hash = "sha256-UsuxILm1G6PkmVw0I/JF12CRltAfCJQFOaT4hFwvR8E=";
+    };
+
+    cargoDeps = rustPlatform.fetchCargoVendor {
+      inherit src;
+      inherit (src) pname version;
+      # hash = "sha256-ElDatyOwdKwHg3bNH/1pcxKI7LXkhsotlDPQjiLHBwA=";
```
Gozala:
```suggestion
```
Nit: Seems like this can go

rust/tonk-worker/src/worker.rs
```diff
@@ -36,7 +36,6 @@ impl TonkServiceWorker {
     /// # Errors
     ///
     /// Returns a `JsError` if the service worker cannot be initialized.
-    #[wasm_bindgen(constructor)]
```
Gozala, line 39: Nit: Can't tell if this was intentional change, or accidental

Gozala (approved):

rust/tonk-worker/src/worker.rs
cdata: It fixes a lint, so intentional and lints will fail if you put it back

PR #373, opened 2026-01-22
feat(worker): introduce identity and space delegations

jackddouglas:
### Description

Introduces user identity, space ownership via UCAN delegations, and persistent key management to tonk-worker. Users now have a unique DID, and spaces are owned via UCAN delegation chains.

- add `Identity` module for user identity management
- add `KeyStore` for persisting keys in IndexedDB
- add `Account` for user preferences (default space, known spaces)
- add `Workspace` abstraction combining user identity with space access
- add `WebCryptoIssuer` adapter for dialog-db's `AsyncIssuer` trait
- add `/api/identify`, `/api/delegations`, `/api/authorize`, `/api/status` endpoints
- add `/api/sync`, `/api/push`, `/api/pull` endpoints for upstream synchronization

### Other changes

### Tested

### Related issues

### Backwards compatibility

### Documentation

---

## PR-Codex overview
This PR focuses on enhancing the `tonk` project's architecture by introducing `Identity` and `Session` management, improving delegation handling, and restructuring API endpoints for better session-based interactions.

### Detailed summary
- Added `Identity` management for user persistence.
- Introduced `Session` to manage space interactions.
- Updated API endpoints to use session-based context.
- Enhanced delegation handling for UCAN authorization.
- Refactored `Cargo.toml` files for dependencies.
- Introduced new response structures for API calls.
- Improved error handling across modules.

> The following files were skipped due to too many changes: `Cargo.lock`


rust/tonk-space/src/space.rs
```diff
@@ -85,12 +87,9 @@ impl<Backend: PlatformBackend + 'static> Space<Backend> {
...
     ) -> Result<Self, SpaceError> {
-        // Open the replica with the operator as issuer and space DID as subject
-        let replica = Replica::open(
-            ReplicaOperator::from(operator),
-            space_did.clone().into(),
-            backend,
-        )?;
+        // Open the replica with the operator and space DID as subject
+        let replica_operator = ReplicaOperator::from(operator);
+        let replica = Replica::open(replica_operator, space_did.clone().into(), backend)?;
```
Gozala, line 111: Nit: would appreciate not having cosmetic changes that have no effect

rust/tonk-space/src/space.rs
```diff
@@ -135,12 +134,9 @@ impl<Backend: PlatformBackend + 'static> Space<Backend> {
...
```
Gozala, line 158: Nit: would appreciate not having cosmetic changes that have no effect, because it takes time to realize they don't have affect.

rust/tonk-space/src/space.rs
```diff
@@ -388,6 +384,53 @@ impl<Backend: PlatformBackend + 'static> Space<Backend> {
...
+        // Query for entities with the specified audience
+        let query = Match::<With<schema::ucan::Audience>> {
+            this: Term::var("delegation"),
+            has: Term::from(audience_did.to_string()),
+        };
+
+        let results: Vec<_> = match query.query(self.clone()).try_collect().await {
+            Ok(r) => r,
+            Err(_) => return vec![],
+        };
```
Gozala, line 429: I don't know if swallowing errors is the best course of action here. Maybe let the error propagate or if it is the best action maybe add code comment explaining why.

rust/tonk-space/src/space.rs
```diff
@@ -388,6 +384,53 @@ impl<Backend: PlatformBackend + 'static> Space<Backend> {
...
+
+        // For each match, fetch the blob and deserialize to Delegation
+        let mut delegations = Vec::new();
+        for result in results {
```
Gozala: Nit: think you can simplify this by leveraging query engine more you can do something like

```rs
#[derive(Concept)]
struct DelegationsByAudience {
   audience: schema::ucan::Audience,
   blob: schema::ucan::Blob
}


let delegations = Match::<DelegationsByAudience> {
    this: Term::var("ucan"),
    audience: Term::from(audience_did.to_string()),
    blob: Term::var("blob")
};

delegations.query(self.clone)
```

Query engine will do the joining and filtering for you

rust/tonk-worker/src/workspace.rs
```diff
@@ -0,0 +1,181 @@
...
+    #[error("Space error: {0}")]
+    Space(#[from] SpaceError),
+
+    /// Failed to create delegation.
+    #[error("Delegation error: {0}")]
+    Delegation(String),
+}
+
+/// An active workspace - a space opened by a specific user.
+pub struct Workspace {
```
Gozala: Nit: Seems like `Session` might be more clear name for this

rust/tonk-worker/src/workspace.rs
```diff
@@ -0,0 +1,181 @@
...
+    /// The user's DID.
+    user_did: String,
```
Gozala: Nit: I would call this account, personally

rust/tonk-worker/src/workspace.rs
```diff
@@ -0,0 +1,181 @@
...
+    /// The opened space.
+    space: Space<ServiceWorkerStorageBackend>,
+    /// The space's operator (for signing space-level operations).
+    space_operator: Operator,
```
Gozala: Nit: I would call this `operator` personally. As in session where operator is authorized to act on behalf of account in specific space.

rust/tonk-worker/src/workspace.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+impl Workspace {
+    /// Open an existing space (or the default space if space_did is None).
```
Gozala: Nit: I have not read through yet and it may become clear, but reading this I'm wondering what does `default` space mean. Perhaps comment could be clear about it.

rust/tonk-worker/src/workspace.rs
```diff
@@ -0,0 +1,181 @@
...
+            Some(did) => did.to_string(),
+            None => identity
+                .account()
+                .default_space()
+                .await?
+                .ok_or(WorkspaceError::NoDefaultSpace)?,
+        };
+
+        // Get the space operator from key store
+        let space_operator = identity
```
Gozala: Nit: I may have wrong mental model here, but in my mind `operator` represents an authority that acts on behalf of account in a specific authorization session. This is to say I find it confusing that we find operators by account.

rust/tonk-worker/src/workspace.rs
```diff
@@ -0,0 +1,181 @@
...
+            .key_store()
+            .space_operator(&space_did)
+            .await?
+            .ok_or_else(|| WorkspaceError::SpaceNotFound(space_did.clone()))?;
+
+        // Open the space database
+        let db_name = format!("tonk-space:{}", space_did);
```
Gozala: ❓ Why prefix db name with `tonk:space:` ? As in it's already a did and I think db's were named after those did's this makes me think this is somehow special, is it ? Or maybe it just uses different names I'm not sure and bit confused.

If it is special maybe we could have some wrapper struct like `struct TonkSpace(Space)` and it could have just `::open() → Self`  doing these things.

If it isn't special, then I'm more confused.

rust/tonk-worker/src/identity.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    /// Failed to access account.
+    #[error("Account error: {0}")]
+    Account(#[from] AccountError),
+}
+
+/// A user's persistent identity on this device.
+///
+/// There is exactly one Identity per device/browser. The identity is backed by
+/// an Ed25519 keypair that is generated on first use and persisted to storage.
```
Gozala, line 28: Nit: This is my mental model of what `Operator` meant to be, So I'm bit confused what the relation to Operator is now or if we need both if two could just merge into one thing.

rust/tonk-worker/src/key_store.rs
```diff
@@ -0,0 +1,297 @@
...
+    /// approach using extractable keys; Plan B will use proper WebCrypto
+    /// non-extractable keys.
+    #[derive(Clone)]
+    pub struct KeyStore {
+        db: Arc<Database>,
+    }
+
+    // SAFETY: Web browsers run Wasm in a single thread.
+    unsafe impl Send for KeyStore {}
+    unsafe impl Sync for KeyStore {}
```
Gozala, line 78: Nit: We probably should consider uplifting these things into dialog-db itself so we don't have to keep doing this.

rust/tonk-worker/src/key_store.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    /// Storage for operator keys in IndexedDB.
+    ///
+    /// Keys are stored as 32-byte secret arrays. This is Plan A's temporary
+    /// approach using extractable keys; Plan B will use proper WebCrypto
+    /// non-extractable keys.
+    #[derive(Clone)]
+    pub struct KeyStore {
+        db: Arc<Database>,
+    }
```
Gozala, line 74: Is there reason not to use existing `IndexedDbStorageBackend` ? It already offers key value abstraction covering bunch of boilerplate done here and you can even customize key / value and their encoding

https://github.com/tonk-labs/tonk/blob/99ecf9d01745f85dfd198eb027bc683a67003864/rust/tonk-worker/src/storage.rs#L41-L42

rust/tonk-worker/src/identity.rs
```diff
@@ -0,0 +1,139 @@
...
+    key_store: KeyStore,
+    account: Account,
+}
+
+impl Identity {
+    /// Load an existing identity or create a new one.
+    ///
+    /// On first call, generates a new random Ed25519 keypair and persists it.
+    /// On subsequent calls, loads the existing keypair from storage.
+    pub async fn load_or_create() -> Result<Self, IdentityError> {
```
Gozala, line 58: Nit: I would love if we started using capability system in new code e.g. here we could have defined something like

```rs
pub struct KeyStore;
impl Attenuation for KeyStore {
  type Of = Subject;
}

pub struct Operator {}
impl Attenuation for Operator {
  type Of = KeyStore;
}


pub struct Load;
impl Effect for Load {
  type Of = Operator;
  type Output = Result<...>;
}

pub struct Create;
impl Effect for Create {
  type Of = Operator;
  type Output = Result<...>;
}
```

Then you'll have effectful method here something like

```rs
pub async fn <Env> (env: &mut Env) → Result<Self, IdentityError> 
where
   Env: Provider<keystore:::Load> + Provider<keystore::Create>
{
   let operator = match keystore::Load.perfom(&mut env).await? {
       Some(op) => opt,
       None => keystore::Create.perform(&mut env).await?
   }

   // ...
}
```

Not suggesting to make this change here, just encouraging it in the future

rust/tonk-worker/src/account.rs
```diff
@@ -0,0 +1,200 @@
...
+}
+
+impl Account {
+    /// Open or create the account for the given user.
+    ///
+    /// # Arguments
+    /// * `user_did` - The user's DID (used to name the database)
+    /// * `operator` - The operator for signing account operations
+    pub async fn open(user_did: &str, operator: &Operator) -> Result<Self, AccountError> {
+        let db_name = format!("tonk-account:{}", user_did);
```
Gozala: Nit: I am confused why do we prefix user did or whether it is necessary, if prefix is needed code comment would be appreciated.

rust/tonk-worker/src/account.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    /// Get the default space DID.
+    ///
+    /// # Returns
+    ///
+    /// - `Ok(Some(did))` if a default space is set
+    /// - `Ok(None)` if no default space has been set
+    /// - `Err(...)` if the query fails
+    pub async fn default_space(&self) -> Result<Option<String>, AccountError> {
```
Gozala: Nit: I find it confusing that account / user would have a default space. Maybe it makes sense, but make me wonder if this is shared state across devices vs local state on device.

Even if shared it also makes me wonder if default per app makes more sense than default per account.

ℹ️ Nothing actionable here, but thought I'd voice this so we can discuss with a wider group.

rust/tonk-worker/src/account.rs
```diff
@@ -0,0 +1,200 @@
...
+//! Unlike secrets (which go in KeyStore), this is non-sensitive metadata
+//! that can be stored as regular dialog-db facts.
+
+use dialog_query::{Attribute, Entity, With};
+use thiserror::Error;
+use tonk_space::{Operator, Space, SpaceError};
+
+use crate::ServiceWorkerStorageBackend;
+
+// Account-specific attributes
```
Gozala: Nit: I you make it `///` comment macro makes use of it and I'd encourage you to do it. In the future I expect dialog capturing attributes and concepts so you can search and discover them.

rust/tonk-worker/src/account.rs
```diff
@@ -0,0 +1,200 @@
...
+#[derive(Attribute, Clone, PartialEq)]
+struct DefaultSpace(pub String);
+
+#[derive(Attribute, Clone, PartialEq)]
```
Gozala, line 20: Nit: Comment explaining what this would be helpful

rust/tonk-worker/src/account.rs
```diff
@@ -0,0 +1,200 @@
...
+        let mut transaction = self.space.edit();
+
+        // Retract any existing default space
+        if let Some(old_default) = self.default_space().await? {
+            tonk_common::log!("Retracting old default space: {}", old_default);
+            transaction.retract(With {
+                this: account_entity.clone(),
+                has: DefaultSpace(old_default),
+            });
+        }
```
Gozala: I'm pretty sure this is redundant, if attribute cardinality isn't may asserting new attribute would succeed prior association. 

P.S: If this isn't working I think we have a bug, and would be good to link to a bug we're working around here.

rust/tonk-worker/src/account.rs
```diff
@@ -0,0 +1,200 @@
...
+        Ok(())
+    }
+
+    /// Get all known space DIDs.
+    ///
+    /// # Returns
+    ///
+    /// - `Ok(vec)` with the list of known space DIDs (may be empty)
+    /// - `Err(...)` if the query fails
+    pub async fn known_spaces(&self) -> Result<Vec<String>, AccountError> {
```
Gozala, line 69: I'm afraid it probably is not going to do what you expect it to. Your attribute needs to have cardinality many to allow multiple associations, which can be done like this

https://github.com/dialog-db/dialog-db/blob/e88a5dce82136459771965be80ed82218a5c3d54/rust/dialog-query-macros/src/lib.rs#L1012

Even then I'm not sure how well we actually support this right now, I suspect it may not work as expected.

rust/tonk-worker/src/account.rs
Gozala: hmm I may be wrong and we don't actually retract prior fact, I'll need to refresh my memory on this.

rust/tonk-worker/src/account.rs
Gozala, line 20:
```suggestion
#[derive(Attribute, Clone, PartialEq)]
#[cardinality(many)]
```

I think you imply cardinality many in which case you should declare it as default is cardinality 1.

rust/tonk-worker/src/account.rs
```diff
@@ -0,0 +1,200 @@
...
+    /// Does nothing if the space is already known.
+    pub async fn add_known_space(&mut self, space_did: &str) -> Result<(), AccountError> {
+        tonk_common::log!("Adding known space: {}", space_did);
+
+        // Check if already known
+        let known = self.known_spaces().await?;
+        if known.contains(&space_did.to_string()) {
+            tonk_common::log!("Space already known, skipping");
+            return Ok(());
+        }
```
Gozala: Nit: You can just assert if association already exists it will be noop, otherwise association will be created.

rust/tonk-worker/src/account.rs
Gozala, line 1: I'm not yet sure how useful it is to capture index known spaces. They way I imagined this working (and was the case in CLI if I recall correctly) known spaces or rather spaces account has access to were just queries for ucans delegations giving account an authority. Same was true for operator, simply find all delegation subjects where audience is operator and that gives you a set of spaces operator has access to.

This both simplifies and complicates things in some sense, it simplifies because you don't need to manage another set of associations, but does require more complicated queries.

rust/tonk-worker/src/identity.rs
```diff
@@ -0,0 +1,139 @@
...
+        let operator = match key_store.user_operator().await? {
+            Some(op) => op,
+            None => key_store.create_user_operator().await?,
+        };
+
+        let account = Account::open(&operator.did().to_string(), &operator).await?;
+
+        Ok(Self {
+            operator,
+            key_store,
```
Gozala, line 72: Nit: It's not clear to me why is this holding onto `key_store` perhaps it should not ? If it does some code comments explaining this would be helpful.

rust/tonk-worker/src/identity.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /// Create a new space owned by this user and open it as a workspace.
+    ///
+    /// The new space will have:
+    /// - A newly generated Ed25519 keypair
+    /// - A delegation granting this user full authority over the space
+    /// - The space set as the default space for this user
+    pub async fn create_workspace(&mut self) -> Result<Workspace, WorkspaceError> {
+        Workspace::create(self).await
```
Gozala: Nit: I think it may make more sense to decouple creation from saving. As in it would make sense to create workspace but have separate `.save(&mut storage)` to persist it as opposed to do it implicitly. I'm also guessing this is why `key_store` is held by the `Identity` and separating these concerns would free it from burden.

rust/tonk-worker/src/identity.rs
```diff
@@ -0,0 +1,139 @@
...
+        let did = identity.did().to_string();
+        assert!(did.starts_with("did:key:z"));
+    }
+
+    #[tokio::test]
+    async fn it_displays_as_did() {
+        let identity = Identity::load_or_create().await.unwrap();
+        let display = format!("{}", identity);
+        assert!(display.starts_with("did:key:z"));
+    }
```
Gozala, line 144: I think it would be good to have test showing that loading across sessions returns same did.

This tests also make me realize that it would be far better to pass storage to `load_or_create` otherwise it's stateful and can lead to difficulties in testing

rust/tonk-worker/src/key_store.rs
```diff
@@ -0,0 +1,297 @@
...
+//!
+//! This module provides `KeyStore`, which manages Ed25519 keys used for signing.
+//!
+//! # Storage Layout
+//!
+//! ```text
+//! IndexedDB: "tonk-keys" (WASM) / In-memory HashMap (native)
+//! └── Object Store: "keys"
+//!     ├── "user"          → { secret: [u8; 32] }
+//!     └── "space:{did}"   → { secret: [u8; 32] }
```
Gozala, line 11: This is more of meta comment than asking to change something here. I do not think we should be holding onto space keys as those are liability. What I think we should be storing in something that in my mind was "credentials" store is [powerline](https://github.com/ucan-wg/delegation?tab=readme-ov-file#powerline)  delegations from  space to owner (account/user) that way they have complete authority over the space without having to manage bunch of keys. Note that unlike keys those delegations can be kept public as those can't be exploited without account keys.

Relatedly what I imagined "authorized session" would be (which is roughly what Workspace is here) is delegation from user account to operator (unique per device). If operator acts with sudo privileges this will be a powerline delegation, otherwise it will be delegation for concrete spaces account gives operator access to.

Gozala (approved):
I think there are couple of things I would like to get addressed, and doing those as followup work seems reasonable to me. Specificaly

1. Dialog already offers abstraction over IDB and it seems like we could just reuse it.
2. I find prefixes on dbs confusing and unnecessary, although I could be missing something there. Specifically I don't think account space should be treated any different from any other space hence require no different prefixing.

Separately  some things feel confusing to me. Specifically I would expect that we load or create operator key first and if it has no authorized session (from an account) we obtain it from account (probably from passkey) gaining access to spaces that account has shared. I think things are either going in different order here or maybe I get tripped by naming.

PR #384, opened 2026-01-30
fix: occasional missing block on /api/sync/pull

Gozala:
There was a race condition in the way upstream branches were configured that caused failing sync behavior in an inconsistent manner.

Issue was fixed in dialog https://github.com/dialog-db/dialog-db/pull/154 and this propagates fix here.

In addition I also added a inspect route `/api/site/:site/:subject/archive/index/:hash` that was helpful in debugging the problem. It allows reading specific block from remote site

---

## PR-Codex overview
This PR introduces new functionality for handling archive blocks in the `tonk-worker` and refines session management in the `tonk-space` module. It also updates dependencies and improves delegation handling in tests.

### Detailed summary
- Added `archive_block` route in `router.rs` for fetching archive blocks.
- Introduced `join_session` method in `identity.rs` for session management.
- Updated `Cargo.toml` to include new dependencies.
- Refactored session creation to use imported operators.
- Enhanced delegation handling in tests for async support.
- Updated `fetch_remote_archive_block` method in `space.rs` for improved remote block fetching.

### DIFF
--- a/rust/tonk-space/src/space.rs
+++ b/rust/tonk-space/src/space.rs
@@ -11,6 +11,7 @@ use dialog_artifacts::{
 use dialog_query::claim::{Transaction, TransactionError};
 use dialog_query::query::Source;
 use dialog_query::{Concept, DeductiveRule, Entity, Session};
+use dialog_storage::StorageBackend;
 use futures_core::Stream;
 use std::sync::Arc;
 use thiserror::Error;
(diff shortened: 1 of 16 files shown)


rust/tonk-space/src/space.rs
```diff
@@ -388,6 +393,46 @@ impl<Backend: PlatformBackend + 'static> Space<Backend> {
...
+        let remote_site = RemoteSite::load(
+            &site.to_string(),
+            replica.issuer().clone(),
+            replica.storage().clone(),
+        )
+        .await?;
+
+        // Get a reference to a branch (we need to open a connection)
+        // Using "main" as default branch to establish the connection
+        let mut remote_branch = remote_site.repository(repo_did.to_string()).branch("main");
```
jackddouglas, line 475: if i understand correctly, we're just using this to establish a connection so we can access the archive? what happens if there's no `main` branch? and the archive is scoped to repository, not branch, so it would be ignored anyway?

jackddouglas (approved):
looks good!

rust/tonk-space/src/space.rs
Gozala: Correct if there in `main` branch or specified `repo_did` I expect `open().await` below to produce an error because it won't be able to read the revision info for that branch.

> and the archive is scoped to repository, not branch, so it would be ignored anyway?

No open will still fail because on open we resolve revision that branch points to. That said you're correct that archive does not care about revision and we could potentially add `.archive().await` to the repository also so you could get handle to it without resolving a branch.

PR #386, opened 2026-01-30
chore: Enforce shared workspace deps

cdata:
This change adds a check (lint) that fails if any Cargo dependencies are **not** inherited from the workspace Cargo.toml. Unfortunately it is unable to check target-specific dependencies, but this is a good start and would have caught several cases where we broke the spirit of the rule.

---

## PR-Codex overview
This PR focuses on updating dependencies to their latest versions, transitioning to a workspace structure for better dependency management, and making adjustments for compatibility with different target architectures.

### Detailed summary
- Changed `sha2` to `sha2_0_10` in `rust/tonk-space/src/secret.rs`.
- Updated `rand` to `rand_0_8` in `rust/tonk-space/src/operator.rs`.
- Modified `Cargo.toml` files to use workspace dependencies for various crates, including `base58`, `tower`, `ed25519-dalek`, `rand`, `thiserror`, `dirs`, `futures-util`, `futures-core`, `tokio`, `ipld-core`, and `serde`.
- Added a new `cargo-enforce-shared-workspace-deps` package in `nix/rust.nix`.
- Adjusted dependencies for `wasm32` architecture in `Cargo.toml` files.
- Updated `tempfile` dependency to use workspace structure.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ anyhow = "1"
 async-trait = "0.1"
 axum = { version = "0.8", default-features = false }
 axum-wasm-macros = "0.1"
+base58 = "0.2"
 base64 = "0.22"
 blake3 = "1.5"
 chrono = { version = "0.4", default-features = false, features = [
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -35,28 +36,35 @@ chrono = { version = "0.4", default-features = false, features = [
...
 getrandom = { version = "0.3", features = ["wasm_js"] }
 hex = "0.4"
+hkdf = "0.12"
 ipld-core = { version = "0.4", features = ["serde"] }
 leptos = "0.8"
 leptos-use = "0.17"
 leptos_router = "0.8"
 port_check = "0.3"
 rand = "0.9"
+rand_0_8 = { package = "rand", version = "0.8" }
```
Gozala, line 54: Should we just update to same rand so we don't have to do this ?

Cargo.toml
```diff
@@ -35,28 +36,35 @@ chrono = { version = "0.4", default-features = false, features = [
...
 rand_core = "0.9"
 reqwest = "0.13"
 serde = "1"
 serde_bytes = "0.11"
 serde_json = "1"
 serde_ipld_dagcbor = "0.6"
 sha2 = "0.11.0-rc.3"
+sha2_0_10 = { package = "sha2", version = "0.10" }
```
Gozala, line 62: Same as above

Gozala (approved):

Cargo.toml
cdata: It's complicated, because other dependencies haven't updated so you get bounds mismatches and other problems. In this case, I think the ed25519 crate has a dependency on legacy rand and the hkdf crate has a dependency on legacy sha2.

So... ideally yes, always try to update. But, this is a reasonable stop-gap until the ecosystem catches up.

Cargo.toml
cdata: See: https://github.com/tonk-labs/tonk/pull/386#discussion_r2748337198

PR #392, opened 2026-02-09
feat(cli): stores data, interfaces with claude

jackddouglas:
### Description

Restructures the tonk CLI around Dialog concepts and instances and adds first-class support for non-interactive/agent-driven usage via a global `--json` flag and non-interactive auth flows

---

## PR-Codex overview
This PR introduces new modules and enhances existing functionalities in the `tonk-cli` project, primarily focusing on JSON output support for various commands and the management of concepts, instances, and spaces.

### Detailed summary
- Added new modules: `concept`, `instance`, `schema`, and `status`.
- Updated `delegation.rs` to replace `println!` with `eprintln!` for saved messages.
- Enhanced `session.rs`, `space.rs`, and `fact.rs` to support optional JSON output.
- Added `batch` operation to handle batch assert/retract from JSON input.
- Introduced command structure for concept management (define, show, extend, delete).
- Improved error handling and user prompts for concept attributes.
- Updated CLI commands to include JSON flag for structured output.

> The following files were skipped due to too many changes: `rust/tonk-cli/src/schema.rs`, `rust/tonk-cli/src/instance.rs`


.claude/commands/tonk.md
```diff
@@ -0,0 +1,168 @@
...
+# → {"ok":true,"id":"did:key:z...","concept":"Task","data":{"title":"Fix login bug","status":"todo","priority":"high"},"created":1739012345}
+
+# From JSON on stdin
+echo '{"title":"Fix login bug","status":"todo"}' | tonk --json create Task --stdin
+
+# From a JSON file
+tonk --json create Task --file task.json
+```
+
+The instance ID (`did:key:z...`) is randomly generated and returned in the response. Store it to reference the instance later.
```
Gozala, line 67: Nit: Unless you're doing something special it's not randomly generated it's derived from from the content and will be hash of of the initial state used as Ed25519 key

.claude/commands/tonk.md
```diff
@@ -0,0 +1,168 @@
...
+
+# Filter by multiple attributes (client-side filtering)
+tonk --json query Task status=todo priority=high
+```
+
+## Instance operations — `tonk show`, `tonk update`, `tonk delete`
+
+```bash
+# Show full details of an instance
+tonk --json show "did:key:z..."
```
Gozala, line 87: Have not scanned through details yet, but I'm bit puzzled in regards to how does it know return `Task` concept. As in entity may have attributes that span across many different concepts or even not fit any concept definition

Gozala: I think thrust of my comment above is I would not recommend putting a different mental model into an agent so it doesn't feel constrained to concept boundaries

.claude/commands/tonk.md
```diff
@@ -0,0 +1,168 @@
...
+# → {"id":"did:key:z...","concept":"Task","data":{"title":"Fix login bug","status":"todo","priority":"high"},"created":1739012345}
+
+# Update specific fields (keys auto-prefixed)
+tonk --json update "did:key:z..." status=done
+# → {"ok":true,"id":"did:key:z...","updated":[{"status":"done"}]}
+
+# Delete an instance
+tonk --json delete "did:key:z..."
+# → {"ok":true,"id":"did:key:z...","concept":"Task"}
```
Gozala, line 96: Nit: Unless you have a good reason to use update / delete terms I would recommend `assert` / `retract` instead as it offers more accurate description. As in you assert a new status on that entity. Or you retract all the known attributes of the entity, it does not actually delete them it just marks those as retracted facts.

We do not expose an API for that yet, but it would be possible to query for retracted facts also.

.claude/commands/tonk.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+## Spaces
+
+```bash
+tonk --json space                    # List accessible spaces
+tonk --json space current            # Show active space
+tonk --json space create "name"      # Create a new space
```
Gozala, line 104: Nit: makes me wonder what happens if one already exists. I personally like following naming convention:

- load: loads if already exists otherwise fails
- create: creates if does not already exists and loads. Fails if already exists
- open: loads if already exists otherwise creates then loads

.claude/commands/tonk.md
```diff
@@ -0,0 +1,168 @@
...
+tonk space set "name-or-did"         # Switch active space
+```
+
+Concepts and instances are scoped to the active space. Switching spaces gives you a different set of concepts and data.
```
Gozala, line 108: I wonder if giving control of creating and switching spaces to agent is the right move as opposed to constraining them to specific space and perhaps even a branch. Alternatively giving authority to create & switch branches could give it probably as much flexibility, yet containment to a specific space.

rust/tonk-cli/src/concept.rs
```diff
@@ -0,0 +1,506 @@
...
+use dialog_query::Value;
+use dialog_query::claim::Attribute;
+use std::str::FromStr;
+
+// ---------------------------------------------------------------------------
+// List all concepts
+// ---------------------------------------------------------------------------
+
+/// List all concepts in the active space.
+pub async fn list(json: bool) -> Result<()> {
```
Gozala, line 22: I think it wold be a lot nicer if context was an argument, it would make testing things a lot simpler.

rust/tonk-cli/src/concept.rs
```diff
@@ -0,0 +1,506 @@
...
+
+// ---------------------------------------------------------------------------
+// Define a new concept
+// ---------------------------------------------------------------------------
+
+/// Define a new concept with the given name and attributes.
+///
+/// Attributes are short names (e.g. `"title"`, `"status"`) that will be
+/// auto-prefixed with the concept namespace (e.g. `"task/title"`).
+pub async fn define(
```
Gozala, line 90: Same as above I would pass ctx as argument

rust/tonk-cli/src/concept.rs
```diff
@@ -0,0 +1,506 @@
...
+    let name = ConceptName::new(name)?;
+
+    let ctx = get_space_context()?;
+    let mut branch = open_branch(&ctx).await?;
+
+    let registry = registry_entity(&ctx.space_did)?;
+    let concept = concept_entity(&ctx.space_did, &name)?;
+
+    // Check if concept already exists
+    let existing_name = fetch_string(&branch, &concept, ATTR_CONCEPT_NAME).await?;
```
Gozala, line 105: Nit: I would say idiomatic way to do this in dialog would be to derive concept in a deterministic way and then assert it. If identical concept exists it would converge and be noop. If concept with different attribute set exists it would produce different concept with perhaps overlapping attributes.

Then problem shifts from there is a conflict to we have a different concept under that same name, do you want to update it or do you want to choose a different name for one of them. If you update you can then assert `the: concept/prior, of: new_concept, is: existist_concept` fact to create a provenance and ability to trace back through the evolutionary chain. You could probably even ask initiator to provide some rational that can be  also captured as a contextual info for the update.

Relatedly I think it is worth exploring if giving unique identifier to agent without creating a human readable name would suffice in some instances. My guess is LLMs don't necessarily care as much about human meaningful names or at least they seem to be much better at seeing difference between two hashes than I am.

rust/tonk-cli/src/concept.rs
```diff
@@ -0,0 +1,506 @@
...
+    // Build instructions
+    let mut instructions = Vec::new();
+
+    // Register concept in registry
+    instructions.push(Instruction::Assert(Artifact {
+        the: Attribute::from_str(ATTR_REGISTRY_CONCEPT)?,
+        of: registry.clone(),
+        is: Value::Entity(concept.clone()),
+        cause: None,
+    }));
```
Gozala, line 143: I think it would be good idea to define a concept for this kind of thing e.g. 

```rs
#[derive(Concept)]
sturct RegisteredConcept {
   /// Concept registry
   this: Entity,
   /// Concept registered
   concept: Entity,
   /// Name under which concept is registered
   name: Name
}
```

That you can just create instance and `.assert` it in a transaction like shown here https://github.com/dialog-db/dialog-db/blob/a242eb30fbede3bff81cf60ad55995cc20fe5d17/rust/dialog-query/README.md#concepts-2

Gozala: Although thinking of it that's probably not how I would model it personally I would just have something like `{ the: 'concept/name', of: concept, is: "Task" }` you can then query just for concepts that have name meaning they are registered.

rust/tonk-cli/src/concept.rs
```diff
@@ -0,0 +1,506 @@
...
+    // Set concept name
+    instructions.push(Instruction::Assert(Artifact {
+        the: Attribute::from_str(ATTR_CONCEPT_NAME)?,
+        of: concept.clone(),
+        is: Value::String(name.to_string()),
+        cause: None,
+    }));
+
+    // Set description if provided
+    if let Some(desc) = &description {
```
Gozala, line 154: I would make description required, as I expect that would greatly improve agents ability to reason about them. At least that has being my experience in the past

rust/tonk-cli/src/concept.rs
```diff
@@ -0,0 +1,506 @@
...
+    description: Option<String>,
+    json: bool,
+) -> Result<()> {
+    let name = ConceptName::new(name)?;
+
+    let ctx = get_space_context()?;
+    let mut branch = open_branch(&ctx).await?;
+
+    let registry = registry_entity(&ctx.space_did)?;
+    let concept = concept_entity(&ctx.space_did, &name)?;
```
Gozala, line 102: ℹ️ Concepts do have unique identifiers that are derived form their definition 

https://github.com/dialog-db/dialog-db/blob/ad4fdf1f28cda3136f99e34062e29ae09bdfba18/rust/dialog-query/src/predicate/concept.rs#L326-L375

Specifically we derive hashes for all the attributes under that concept and then hash that set. What that means is that field names no longer matter neither does a concept name. It somewhat similar to [a big idea in unison](https://www.unison-lang.org/docs/the-big-idea/).

That is to say I think you could construct a concept here from the attributes and then use it's `.operator()` as an entity identifier.

rust/tonk-cli/src/concept.rs
```diff
@@ -0,0 +1,506 @@
...
+            cause: None,
+        }));
+    }
+
+    // Add each attribute
+    for attr in &qualified_attrs {
+        instructions.push(Instruction::Assert(Artifact {
+            the: Attribute::from_str(ATTR_CONCEPT_ATTRIBUTE)?,
+            of: concept.clone(),
+            is: Value::String(attr.clone()),
```
Gozala, line 168: Note that just like concepts have unique identifiers based on their composition [so do attributes](https://github.com/dialog-db/dialog-db/blob/ad4fdf1f28cda3136f99e34062e29ae09bdfba18/rust/dialog-query/src/predicate/concept.rs#L354) so instead of this I'd be associating entity identifier with corresponding attribute identifier via some attribute like `ATTR_CONCEPT_ATTRIBUTE`.

I would also separately associate attribute entity with it's [schema properties](https://github.com/dialog-db/dialog-db/blob/ad4fdf1f28cda3136f99e34062e29ae09bdfba18/rust/dialog-query/src/attribute.rs#L23-L27)

It also might be useful to just assert serialized schema as a single fact to avoid complicated queries, but I would probably pun on that until it proves to be a case. But what I'm implying is something along the lines of `{ the: concept/source, of: concept.operator(), is: concept.serialize(ipld) }`

rust/tonk-cli/src/concept.rs
```diff
@@ -0,0 +1,506 @@
...
+// ---------------------------------------------------------------------------
+// Show concept details
+// ---------------------------------------------------------------------------
+
+/// Show the schema of a concept.
+pub async fn show(name: String, json: bool) -> Result<()> {
+    let ctx = get_space_context()?;
+    let branch = open_branch(&ctx).await?;
+
+    let name = ConceptName::new(name)?;
```
Gozala, line 209: I think if you were to define concept of the concept as a concept you could simply query by something like

```rs
let selector = Match::<RegisteredConcept> { name: name.into(), ... }
let matched = selector.query(...).await?;
```

rust/tonk-cli/src/concept.rs
```diff
@@ -0,0 +1,506 @@
...
+
+    Ok(())
+}
+
+// ---------------------------------------------------------------------------
+// Extend a concept with new attributes
+// ---------------------------------------------------------------------------
+
+/// Add attributes to an existing concept.
+pub async fn extend(name: String, attributes: Vec<String>, json: bool) -> Result<()> {
```
Gozala, line 256: Not going to repeat bunch of things from above, but I would not mutate concept in place would derive new concept assert it and then update name to point to new one also assert relation between new old for provenance.

rust/tonk-cli/src/concept.rs
```diff
@@ -0,0 +1,506 @@
...
+
+    Ok(())
+}
+
+// ---------------------------------------------------------------------------
+// Delete a concept
+// ---------------------------------------------------------------------------
+
+/// Delete a concept and optionally its instances.
+pub async fn delete(name: String, force: bool, json: bool) -> Result<()> {
```
Gozala, line 341: I think deleting is not ideal, if you follow above suggestions deletion (or retraction rather) would simply revoke fact that gives concept a known name. Which means it would not be listed under it but still usable and perhaps we could even allow searching for concepts without given names with some additional argument

rust/tonk-cli/src/instance.rs
```diff
@@ -0,0 +1,625 @@
...
+                    .map(|a| short_attribute(&stored_name, a))
+                    .collect::<Vec<_>>()
+                    .join(", ")
+            );
+        }
+        qualified_fields.push((qualified, value.clone()));
+    }
+
+    // Generate a new random entity for the instance
+    let instance_entity = Entity::new().context("Failed to generate instance entity")?;
```
Gozala, line 89: Provided some notes on this in above files, I think generating unique entities is not ideal, because two actors doing same exact thing will produce different ones. Instead leveraging derived identifiers where possible would be highly encouraged. In this instance we could leverage identifiers that concepts and attributes derive already.

rust/tonk-cli/src/schema.rs
```diff
@@ -0,0 +1,476 @@
...
+// ConceptName newtype
+// ---------------------------------------------------------------------------
+
+/// A validated concept name (alphanumeric, hyphens, underscores only).
+///
+/// Concept names are used as attribute namespace prefixes (e.g. `"task/title"`),
+/// so they must only contain safe characters. This type guarantees that
+/// invariant at construction time.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct ConceptName(String);
```
Gozala, line 54: This probably could be have `derive(Attribute)` see https://github.com/dialog-db/dialog-db/blob/a242eb30fbede3bff81cf60ad55995cc20fe5d17/rust/dialog-query/README.md#attributes for details

rust/tonk-cli/src/schema.rs
```diff
@@ -0,0 +1,476 @@
...
+        &self.0
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Well-known attribute constants
+// ---------------------------------------------------------------------------
+
+/// Registry attribute: points from registry entity to concept entities.
+pub const ATTR_REGISTRY_CONCEPT: &str = "registry/concept";
```
Gozala, line 107: It is not obvious to me what is the value of this unless there are multiple registries, because you could simply query for concepts that have names or something along those lines. Specifically I would have defined concept of `KnownConcept` and `NamedConcept` where later simply extends former with `name` attribute.

rust/tonk-cli/src/schema.rs
```diff
@@ -0,0 +1,476 @@
...
+pub const ATTR_CONCEPT_DESCRIPTION: &str = "concept/description";
+
+/// Concept attribute: one per attribute the concept has (multi-valued).
+pub const ATTR_CONCEPT_ATTRIBUTE: &str = "concept/attribute";
+
+/// Concept attribute: back-reference to instances (multi-valued).
+pub const ATTR_CONCEPT_INSTANCE: &str = "concept/instance";
+
+/// Instance attribute: type reference back to concept entity.
+pub const ATTR_INSTANCE_TYPE: &str = "instance/type";
```
Gozala, line 122: These seem like same to me except of direction. But with datalog you can go either direction so one of them seems redundant

rust/tonk-cli/src/schema.rs
```diff
@@ -0,0 +1,476 @@
...
+// Attribute name helpers
+// ---------------------------------------------------------------------------
+
+/// Given a concept name and user-supplied attribute key, produce the fully
+/// qualified attribute name `{concept_lower}/{key}`.
+///
+/// If the key already contains a `/` and the prefix matches the concept
+/// namespace, it is returned as-is. If the prefix doesn't match, an error
+/// is returned.
+pub fn qualify_attribute(concept_name: &ConceptName, key: &str) -> Result<String> {
```
Gozala, line 186: While I think it is generally a good idea, I do think that limiting to attribute namespace to concept name is an anti pattern. This is something we learned from experience before we transitioned to ECS style query engine. Also something that datomic community is very vocal about.

Gozala (approved):
I have provided bunch of feedback in regards to how I think this could be improved, but I think we should consider doing those improvements as followups over time.

A thing that worries me however is that lot of functionality is added but nothing is tested which means it would be easy to break things and not know about it. Please add at least some basic testing so at least happy paths can remain functional as we continue develop ment. Given a large stack of changes I'm ok with this as followup.

P.S.: Lack of tests means that after propagating dialog / ucan changes I either hope that type checker catches all problems or I have to manually try each function & I would very much prefer if I could instead run tests to verify

jackddouglas:
i have integration tests queued up here: https://github.com/tonk-labs/tonk/pull/401

.claude/commands/tonk.md
jackddouglas: the trick here is we currently allow space name conflicts since we generate a new DID for each space and index on that. so if you create a new space with the same name as an existing one, confusingly it won't err and will instead redirect you to a new space named the same as an existing one.

so we need to decide where to enforce uniqueness. for now i think i'll make it so space names must be unique per authority, since we shouldn't have name conflicts for spaces owned by different users

.claude/commands/tonk.md
jackddouglas: ah okay i assumed it was random because of this in dialog's `entity.rs`:

```rust
impl Entity {
    /// Initialize a new [`Entity`] with a randomly generated, globally unique
    /// URI. The URI is formatted as an ed25519 DID Key.
    pub fn new() -> Result<Entity, DialogArtifactsError> {
        Self::try_from(Uri::unique()?)
    }

    /// Get the [`Entity`] as a string reference
    pub fn as_str(&self) -> &str {
        (**self).as_str()
    }

    /// Get the raw byte representation of the [`Entity`] as it should be
    /// formatted for use in an index key.
    pub fn key_bytes(&self) -> &[u8; ENTITY_LENGTH] {
        &self.1
    }
}
```

will add a helper in tonk that derives from field content though

PR #394, opened 2026-02-10
feat(cli): TON-1858: rules

jackddouglas:
### Description

Adds support for Dialog's deductive rule system.

When querying a concept that has associated rules, derived instances are transparently merged with stored instances via dialog-db's Session-based querying.

New CLI commands:
- `tonk rule`: list all rules
- `tonk rule define <name> --file <json> | --stdin`: define a rule from JSON
- `tonk rule show <name>`: show rule details
- `tonk rule delete <name>`: delete a rule

Details:
- rules are stored as JSON blobs in dialog-db with `rule/name`, `rule/conclusion`, `rule/definition`, and `rule/description` attributes
- rule entities are deterministically derived from `space_did + rule_name` (same pattern as concepts)
- at query time, rules targeting the queried concept are loaded, compiled into `DeductiveRule`s, registered with a `Session`, and executed via `Concept::apply`
- compiler automatically renames user variables to match concept operand names, with collision detection and auto-renaming for conflicting variables
- validation catches errors early: missing concepts, unbound variables, invalid attribute references

### Related issues

- closes TON-1858

---

## PR-Codex overview
This PR introduces a new module for managing rules in the `tonk-cli`, enabling the definition, listing, showing, and deletion of deductive rules. It enhances the schema and query functionalities to incorporate rules, allowing for more complex data derivation.

### Detailed summary
- Added `pub mod rule;` to `lib.rs`.
- Introduced new attributes related to rules in `schema.rs`.
- Implemented `rule_entity` function to derive rule entities.
- Added `RuleCommands` enum for rule management in the CLI.
- Updated command handling in `tonk.rs` to include rule commands.
- Enhanced `query` function to support selectors instead of filters.
- Created functions for defining, showing, listing, and deleting rules in `rule.rs`.
- Validated rule definitions to ensure they meet schema requirements.
- Added JSON serialization for rule definitions and responses.

### DIFF
--- a/.claude/commands/tonk.md
+++ b/.claude/commands/tonk.md
@@ -107,6 +107,93 @@ tonk space set "name-or-did"         # Switch active space
 
 Concepts and instances are scoped to the active space. Switching spaces gives you a different set of concepts and data.
 
+## Rules — deriving instances from patterns
+
+A rule defines how instances of a concept can be **derived** from patterns across existing facts. Rules use Datalog-style deduction: positive premises (`when`) that must hold, and negative premises (`unless`) that must NOT hold. Variables (prefixed with `?`) bind values and create implicit joins across premises.
+
+Multiple rules can derive the same concept — they act as **OR branches**. Each rule independently produces derived instances, and results are merged. This lets you evolve rules independently rather than maintaining one giant rule with many branches.
+
+When you query a concept that has rules, derived instances are transparently merged with stored instances.
+
+```bash
+# List all rules
+tonk --json rule
+# → [{"name":"safe-meals","conclusion":"SafeMeal"},...]
+
+# Define a rule from JSON (via file)
+tonk --json rule define safe-meals --file rule.json
+
+# Define a rule from JSON (via stdin)
+cat <<'EOF' | tonk --json rule define safe-meals --stdin
+{
+  "conclusion": {
+    "concept": "SafeMeal",
+    "this": "?allergy",
+    "bindings": {
+      "attendee": "?person",
+      "recipe": "?recipe_name"
+    }
+  },
+  "when": [
+    { "the": "allergy/person", "of": "?allergy", "is": "?person" },
+    { "the": "recipe/name", "of": "?recipe", "is": "?recipe_name" },
+    { "the": "recipe/ingredient", "of": "?recipe", "is": "?ingredient" }
+  ],
+  "unless": [
+    { "the": "allergy/substance", "of": "?allergy", "is": "?ingredient" }
+  ]
+}
+EOF
+# → {"ok":true,"name":"safe-meals","conclusion":"SafeMeal","when_count":3,"unless_count":1}
+
+# Show rule details
+tonk --json rule show safe-meals
+
+# Delete a rule
+tonk --json rule delete safe-meals
+
+# Query now includes derived instances transparently
+tonk --json query SafeMeal
+```
+
+### Rule definition JSON format
+
+```json
+{
+  "conclusion": {
+    "concept": "ConceptName",
+    "this": "?entity_var",
+    "bindings": { "attr_short_name": "?variable", ... }
+  },
+  "when": [
+    { "the": "namespace/attribute", "of": "?var_or_constant", "is": "?var_or_constant" },
+    ...
+  ],
+  "unless": [
+    { "the": "namespace/attribute", "of": "?var_or_constant", "is": "?var_or_constant" },
+    ...
+  ]
+}
+```
+
+**Conclusion fields:**
+- `concept` — the name of the concept being derived (must already be defined; referencing an undefined concept is an error)
+- `this` — which entity variable from the premises becomes the derived entity's identity. `this` is distinct from attribute bindings: it designates the *entity identity* of each derived instance, not an attribute value. If omitted, defaults to the first entity variable in the `when` premises.
+- `bindings` — maps concept attribute short names to premise variables (`"?var"`) or constant values (any string without `?` prefix). Not every premise variable needs to appear in bindings — variables can appear only in premises to serve as join variables (e.g. `?ingredient` in the example above joins `when` and `unless` premises without appearing in the conclusion). The rule compiler automatically renames variables to match the concept's operand names; if this would cause a collision with another variable, the conflicting variable is auto-renamed.
+
+**Premise terms:**
+- `"?name"` — variable (binds/joins by name across premises)
+- `"_"` — wildcard (matches anything, no binding)
+- any other string — constant (exact match filter)
+
+**Constraints:**
+- `the` is always a fully qualified attribute name (constant). Premise attributes reference existing data but are not validated against concept schemas — only the conclusion concept must be defined.
+- Every variable in conclusion `bindings` must appear in at least one positive (`when`) premise (appearing only in `unless` is not sufficient — this is a Datalog safety requirement)
+- The conclusion concept must already be defined
+- The `unless` section is optional (negation-as-failure)
+- Rule names are case-insensitive (stored and looked up in lowercase). Names serve as human-friendly identifiers for lookup and deletion.
+- Rules must have at least one positive premise in `when`
+
 ## Typical workflow
 
 ```bash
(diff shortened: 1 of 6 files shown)


.claude/commands/tonk.md
```diff
@@ -107,6 +107,91 @@ tonk space set "name-or-did"         # Switch active space
...
+# Define a rule from JSON (via stdin)
+cat <<'EOF' | tonk --json rule define safe-meals --stdin
+{
+  "conclusion": {
+    "concept": "SafeMeal",
+    "this": "?allergy",
+    "bindings": {
+      "attendee": "?person",
+      "recipe": "?recipe_name"
+    }
```
Gozala, line 135: How is `this` different from other bindings ? And if it is not, why break it out separately ? Not implying we do this this but I have personally found following form to be most intuitive 

```js
{
  assert: {
    SafeMeal: {
      this: ?allergy,
      attendee: ?person,
      recipe: ?recipe_name
    },
  },
  when: [..],
  unless: [...] 
}
```

.claude/commands/tonk.md
```diff
@@ -107,6 +107,91 @@ tonk space set "name-or-did"         # Switch active space
...
+- `"_"` — wildcard (matches anything, no binding)
+- any other string — constant (exact match filter)
+
+**Constraints:**
+- `the` is always a fully qualified attribute name (constant)
+- Every variable in conclusion `bindings` must appear in at least one `when`/`unless` premise
+- The conclusion concept must already be defined
+- The `unless` section is optional (negation-as-failure)
+- Rule names are case-insensitive (stored and looked up in lowercase)
+- Rules must have at least one positive premise in `when`
```
Gozala, line 195: Every variable MUST appear in one ore more premises inside `when`.

.claude/commands/tonk.md
```diff
@@ -107,6 +107,91 @@ tonk space set "name-or-did"         # Switch active space
...
+tonk --json rule show safe-meals
+
+# Delete a rule
+tonk --json rule delete safe-meals
+
+# Query now includes derived instances transparently
+tonk --json query SafeMeal
+```
+
+### Rule definition JSON format
```
Gozala, line 159: I'm not sure if every variable is expected to be listed in the bindings or if they could appear in the premises regardless of being in the bindings.

.claude/commands/tonk.md
Gozala, line 1: From reading this it's not obvious to me what happens if rule refers to a concept that has not being defined, would be be great if document covered that part

.claude/commands/tonk.md
```diff
@@ -107,6 +107,91 @@ tonk space set "name-or-did"         # Switch active space
...
+
+When you query a concept that has rules, derived instances are transparently merged with stored instances.
+
+```bash
+# List all rules
+tonk --json rule
+# → [{"name":"safe-meals","conclusion":"SafeMeal"},...]
+
+# Define a rule from JSON (via file)
+tonk --json rule define safe-meals --file rule.json
```
Gozala, line 124: I think it should be made clear that rules do not need names and are there purely for convenience as in people can refer them and look them up by it, but probably something that is optional.

General thinking is that rules are just or branches for concept derivation. It is just convenient not to have to have giant rule with many or branches and instead have bunch of small ones that can be evolved independently.

rust/tonk-cli/src/instance.rs
```diff
@@ -188,34 +193,63 @@ pub async fn query(concept_name: String, filters: Vec<String>, json: bool) -> Re
         qualified_filters.push((qualified, parse_value(value)));
     }
 
-    // Get all instance entities for this concept
-    let instance_entities = fetch_entity_values(&branch, &concept, ATTR_CONCEPT_INSTANCE).await?;
+    // Check for rules targeting this concept
+    let rule_defs =
+        crate::rule::load_rules_for_concept(&branch, &ctx.space_did, &stored_name).await?;
+
+    let rows = if !rule_defs.is_empty() {
```
Gozala: It's not obvious to me that branching between these cases, intuitively it would be simpler to just have one code path with potentially empty list list of rules.

rust/tonk-cli/src/rule.rs
```diff
@@ -0,0 +1,850 @@
...
+
+    /// Maps concept attribute short names to variables.
+    /// e.g. `{"attendee": "?person", "recipe": "?recipe"}`
+    pub bindings: HashMap<String, String>,
+
+    /// Which entity variable in the premises should bind to the
+    /// concept's implicit `this` entity. e.g. `"?allergy"`.
+    ///
+    /// If omitted, defaults to the first entity variable (`of` field)
+    /// found in the `when` premises.
```
Gozala, line 84: I'm not sure I'm fan of this default, personally. I'd error on the side being explicit or convention like `?this` being the implicit name.

Gozala (commented):
I think there is some overlap with what query engine and things could probably be simplified some if we take advantage of it. Rules probably need to support more kinds of premises, but all that is probably things that can be improved incrementally (in followups).

rust/tonk-cli/src/instance.rs
```diff
@@ -228,20 +262,122 @@ pub async fn query(concept_name: String, filters: Vec<String>, json: bool) -> Re
...
+}
+
+/// Session-based query with rules. Compiles rules into DeductiveRules,
+/// registers them with a Session, and uses dialog-db's concept application
+/// to merge stored and derived instances.
+async fn query_with_rules(
+    branch: dialog_artifacts::repository::Branch<tonk_space::FsBackend>,
+    schema_attrs: &[String],
+    stored_name: &ConceptName,
+    qualified_filters: &[(String, Value)],
```
Gozala: Without outside context it' really hard for me to tell what this parameter is, would be nice if doc comments provided some insight perhaps even an example

rust/tonk-cli/src/instance.rs
```diff
@@ -188,34 +193,63 @@ pub async fn query(concept_name: String, filters: Vec<String>, json: bool) -> Re
...
+        // No rules — use direct enumeration (existing fast path)
+        query_direct(
+            &branch,
+            &concept,
+            &schema_attrs,
+            &stored_name,
+            &qualified_filters,
+        )
+        .await?
+    };
```
Gozala: This seems like premature optimization. I would much rather have one code path and find ways to optimize it if needed. Rule lookups should be very fast as they're in hash map so I'd be surprised if this fast path even has a meaningful impact

rust/tonk-cli/src/instance.rs
```diff
@@ -166,6 +166,11 @@ pub async fn create(
 /// Query instances of a concept, with optional filters.
 ///
 /// Filters are `key=value` pairs for exact matching.
+///
+/// When rules exist for the concept, uses dialog-db's Session-based
+/// concept querying which automatically merges stored instances with
+/// rule-derived instances. Falls back to direct enumeration when no
+/// rules apply.
 pub async fn query(concept_name: String, filters: Vec<String>, json: bool) -> Result<()> {
```
Gozala: I was getting confused as I read through what the filters were, I think calling it a `selector` would hav made it more legible.

rust/tonk-cli/src/instance.rs
```diff
@@ -228,20 +262,122 @@ pub async fn query(concept_name: String, filters: Vec<String>, json: bool) -> Re
...
+    let dynamic_concept = build_dynamic_concept(schema_attrs)?;
+
+    // Build query parameters: named variables for each attribute
+    // plus "this" for the entity
+    let mut params = Parameters::new();
+    let this_var: Term<Value> = Term::var("this");
+    params.insert("this".to_string(), this_var.clone());
+
+    // Create a variable for each attribute
+    let mut attr_vars: Vec<(String, String, Term<Value>)> = Vec::new(); // (qualified, short, term)
```
Gozala: I think struct with named fields would have made it easier to understand diff between Left and Middle string. After going back and forth I think it's `attribute_name, field_name, term` or maybe `param_name` instead of `field_name` would be a lot more clearer.

rust/tonk-cli/src/instance.rs
```diff
@@ -228,20 +262,122 @@ pub async fn query(concept_name: String, filters: Vec<String>, json: bool) -> Re
...
+        .await
+        .map_err(|e| anyhow::anyhow!("Query failed: {}", e))?;
+
+    // Convert answers to rows
+    let mut rows = Vec::new();
+    for answer in &answers {
+        let mut data = serde_json::Map::new();
+
+        // Resolve the entity (this)
+        let entity_str = match answer.resolve(&this_var) {
```
Gozala, line 293: Sounds like we need `dynamic_concept.realize` similar to our static case so there would be no need to do this manually.

rust/tonk-cli/src/schema.rs
```diff
@@ -127,6 +127,21 @@ pub const ATTR_INSTANCE_CREATED: &str = "instance/created";
...
+
+/// Rule attribute: optional description.
+pub const ATTR_RULE_DESCRIPTION: &str = "rule/description";
+
+/// Rule attribute: name of the conclusion concept.
+pub const ATTR_RULE_CONCLUSION: &str = "rule/conclusion";
+
+/// Rule attribute: JSON-serialized rule definition.
+pub const ATTR_RULE_DEFINITION: &str = "rule/definition";
+
```
Gozala, line 144: I think I mentioned this before in different PRs but I think instead of manual attributes and manual queries we should leverage Concept to represent registered concepts and rules and leverage query engine instead of doing bunch of it's work manually.

Gozala (approved):
Can you please add some tests before landing this so we have some way of knowing that we aren't breaking things as we iterate. All my other comments are advisory and things that are better addressed as followups.

PR #397, opened 2026-02-11
feat(cli): TON-1860: import concepts/rules from yaml

jackddouglas:
## Summary

- adds `tonk import <file.yaml>` CLI command that auto-detects whether a YAML file contains **concept definitions** or **rule definitions** and imports them atomically into the active space
- supports a `--force` flag to overwrite existing concepts/rules instead of failing on collisions
- includes full validation (duplicate detection, schema checks, trial-compilation for rules) before committing any changes

### Concept YAML format

Concepts are defined under namespace keys with a declarative schema:

```yaml
diy.cook:
  Recipe:
    this:
      the: Meal recipe
    title:
      the: The name of this recipe
      as: Text
    ingredient:
      the: Ingredients of the recipe
      cardinality: many
      as: RecipeStep
```

Supports typed attributes (`as: Text`, `as: Integer`, concept references like `as: RecipeStep`), enum constraints (`as: [tsp, mls]`), multi-valued attributes (`cardinality: many`), and optional fields (`optional:` key presence).

### Rule YAML format

Rules use `assert`/`when`/`unless` blocks referencing namespaced concepts:

```yaml
user.rules:
  find-allergy-conflicts:
    assert:
      diy.planner/AllergyConflict:
        person: ?person
        recipe: ?recipe
    when:
      - diy.cook/Recipe:
          this: ?recipe
          ingredient: ?ingredient
      - diy.cook/Ingredient:
          this: ?ingredient
          name: ?substance
      - diy.health/Allergy:
          this: ?allergy
          person: ?person
          substance: ?substance
```

Rules are lowered from concept-oriented YAML into internal EAV triples, validated against the existing concept schema, and trial-compiled before committing.

### Other changes

- adds `serde_yaml` dependency to the workspace
- refactors `rule.rs` to expose `RuleDefinition`, `RuleConclusion`, `RulePremise`, `validate_definition`, and `compile_rule` for reuse by the import module
- updates `instance.rs` to support the new schema constants
- adds example YAML files (`cook.yaml`, `minimal.yaml`, `planner.yaml`, `rules.yaml`) demonstrating the format

## Tested

- unit tests covering: basic concept parsing, enum types, optional attributes, multiple namespaces, empty YAML, minimal concepts, concept reference types, duplicate concept detection, rule parsing/lowering, rule validation, file type auto-detection, and edge cases
- all tests are in `import.rs` and can be run with `cargo test -p tonk-cli`

## Related issues

- closes TON-1860

## Backwards compatibility

Yes, just introduces a new `import` subcommand and new module. No existing commands or data structures are modified in a breaking way.

---

## PR-Codex overview
This PR introduces batch operations for managing instances of concepts in the `tonk` CLI, alongside enhancements to YAML import functionality and schema definitions.

### Detailed summary
- Added `batch` module for creating, updating, and deleting multiple instances atomically.
- Introduced new commands: `batch create`, `batch update`, `batch delete`.
- Enhanced YAML import to support concepts and rules.
- Updated `Cargo.toml` to include `serde_yaml`.
- Added new attributes and metadata for concepts in `schema.rs`.
- Improved error handling and validation for batch operations.

> The following files were skipped due to too many changes: `rust/tonk-cli/src/import/commit.rs`, `rust/tonk-cli/src/import/rule_parse.rs`


rust/tonk-cli/examples/cook.yaml
```diff
@@ -0,0 +1,38 @@
+diy.cook:
+  Recipe:
+    this:
+      the: Meal recipe
+    title:
+      the: The name of this recipe
+      as: Text
+    ingredient:
```
Gozala, line 8: What's the implicit type ? Is it Text, Any or something else ?

rust/tonk-cli/examples/cook.yaml
```diff
@@ -0,0 +1,38 @@
...
+      the: The unit of measurement for the ingredient
+      as: [tsp, mls]
+
+  RecipeStep:
+    this:
+      the: The cooking step
+    after:
+      the: Step to perform this after
+      as: RecipeStep
+      optional:
```
Gozala, line 35: Is this valid yml what's the value ?

rust/tonk-cli/src/import/rule_parse.rs
```diff
@@ -0,0 +1,696 @@
...
+
+// ---------------------------------------------------------------------------
+// Rule YAML parsing
+// ---------------------------------------------------------------------------
+
+/// Extract the concept name from a namespaced reference like `"diy.cook/Recipe"`.
+///
+/// Splits on the last `/` and returns the part after it. If there's no `/`,
+/// returns the whole string.
+fn extract_concept_name(reference: &str) -> String {
```
Gozala, line 81: I think stripping out namespace is a bad idea. Intention that was

1. To communicated namespace attributes to use on concept attributes (as opposed to using just concept name).
2. To allow having more than one definitions of `Recipe` etc.

Gozala (approved):
I have skimmed through and I think it looks good enough. Few things I'd recommend addressing here or as followups

1. Namespace for concepts were implied to be inherited by attributes, which I still think is a good idea.
2. I think we should allow same named concepts to co-exist in different namespaces because names don't really matter and avoiding conflicts seems like a good idea. e.g. if user does not qualify which Recipe and system is aware of more than one it could simply ask or surface both. _(🤔 In fact having two different Recipe concepts is probably a good signal that we should have rules that can bridge them)_.
3. I think it would be best to avoiding generating rule and concept identifiers and instead lean into identifiers derived by dialog which I pointed out in of the other reviews

rust/tonk-cli/examples/cook.yaml
jackddouglas: it's valid, `optional` is parsed as null

rust/tonk-cli/examples/cook.yaml
jackddouglas: there's no recorded type. though at the moment the only difference is in what metadata's recorded, since we don't have enforced typing yet

PR #398, opened 2026-02-11
feat: Universal content-addressed blob storage

cdata:
This change implements universal (as in, platform agnostic) content-addressed blob storage via a `BlobStorage` construct. It supports two operations: read by `Blake3Hash`, and write bytes (receiving the final `Blake3Hash`). It strives to be as efficient as possible, utilizing conventional async IO interfaces and enabling reads and writes without requiring that the whole file be held in memory.

On the web, `BlobStorage` uses the [origin-private file system](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system). This was mostly painless to use, except: OPFS does not have a standard atomic "move," (which the current implementation relies on to enable hashing without holding the whole file in memory). Thankfully, [Chrome has a non-standard `move` that the WG is attempting to standardize](https://github.com/whatwg/fs/pull/180). So on Chromium browsers, we have the fast move and on other browsers we have a fallback that implements a comparatively glacial read-and-pipe-to-writer scheme. It's probably still better than reading a very large file into memory, though.

---

## PR-Codex overview
This PR introduces a new module `tonk-blobs` for local blob storage, enhances the `tonk-cli` with improved command structures, and updates dependencies, particularly for WebAssembly support.

### Detailed summary
- Added `tonk-blobs` module for filesystem-backed blob storage.
- Introduced `TonkBlobsError` enum for error handling in blob operations.
- Updated `Cargo.toml` to include `tonk-blobs` and adjust dependencies.
- Enhanced `tonk-cli` command structure for better command handling.
- Improved WebAssembly support with updated `rustflags`.
- Added tests for blob storage functionality and VFS operations.

> The following files were skipped due to too many changes: `rust/tonk-blobs/src/vfs/web.rs`

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,3 +1,3 @@
 [target.wasm32-unknown-unknown]
 runner = "wasm-bindgen-test-runner"
-rustflags = ['--cfg', 'getrandom_backend="wasm_js"']
+rustflags = ['--cfg', 'getrandom_backend="wasm_js"', '--cfg', 'web_sys_unstable_apis']
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 
 members = [
     "rust/tonk-access-service",
+    "rust/tonk-blobs",
     "rust/tonk-cli",
     "rust/tonk-core",
     "rust/tonk-common",
@@ -50,7 +51,8 @@ futures = "0.3"
 futures-core = "0.3"
 futures-util = "0.3"
 getrandom_0_2 = { package = "getrandom", version = "0.2", features = ["js"] }
-getrandom = { version = "0.3", features = ["wasm_js"] }
+getrandom_0_3 = { version = "0.3", features = ["wasm_js"] }
+getrandom = { version = "0.4", features = ["wasm_js"] }
 hex = "0.4"
 hkdf = "0.12"
 idb = "0.6"
@@ -73,6 +75,8 @@ tempfile = "3"
 thirtyfour = "0.36"
 thiserror = "2"
 tokio = "1"
+tokio-stream = "0.1"
+tokio-util = { version = "0.7", features = ["io"] }
 tower = "0.5"
 tower-service = "0.3"
 hyper = "1"
@@ -80,6 +84,7 @@ hyper-util = "0.1"
 bytes = "1"
 http-body-util = "0.1"
 url = "2"
+ulid = "1"
 wasm-streams = "0.4"
 web-time = "1"
 webbrowser = "1.0"
@@ -97,7 +102,7 @@ web-sys = { version = "0.3" }
 dialog-storage = { git = "https://github.com/dialog-db/dialog-db.git", branch = "tonk" }
 dialog-query = { git = "https://github.com/dialog-db/dialog-db.git", branch = "tonk" }
 dialog-artifacts = { git = "https://github.com/dialog-db/dialog-db.git", branch = "tonk" }
-dialog-common = { git = "https://github.com/dialog-db/dialog-db.git", branch = "tonk" }
+dialog-common = { git = "https://github.com/dialog-db/dialog-db.git",  branch = "tonk" }
 dialog-capability = { git = "https://github.com/dialog-db/dialog-db.git", branch = "tonk" }
 dialog-s3-credentials = { git = "https://github.com/dialog-db/dialog-db.git", branch = "tonk", features = ["ucan"] }
 
--- a/rust/tonk-blobs/src/blob_storage.rs
+++ b/rust/tonk-blobs/src/blob_storage.rs
@@ -1,0 +1,377 @@
+use base58::ToBase58;
+use dialog_common::Blake3Hash;
+use futures_core::Stream;
+use futures_util::StreamExt;
+use tokio::io::AsyncWriteExt;
+
+use crate::{FileReader, TonkBlobsError, Vfs};
+
+/// Derives the content-addressed path for a blob from its Blake3 hash.
+fn hash_to_path(hash: &Blake3Hash) -> String {
+    let b58 = hash.as_bytes().to_base58();
+    format!("{}/{}/{}/{}", &b58[0..2], &b58[2..4], &b58[4..6], b58)
+}
+
+/// High-level abstraction over platform-specific content-addressed blob
+/// storage.
+///
+/// Blobs are stored in a sharded directory tree derived from the base58
+/// encoding of their hash. Given a hash that encodes to `AbCdEfGh...`:
+///
+/// ```text
+/// <vfs root>/
+///   Ab/
+///     Cd/
+///       Ef/
+///         AbCdEfGh...   <- full base58 hash as filename
+/// ```
+///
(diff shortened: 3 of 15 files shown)


Cargo.toml
```diff
@@ -121,3 +126,7 @@ debug = 0
 inherits = "release"
 strip = "debuginfo"
 
+[patch.'https://github.com/tonk-labs/rs-ucan.git']
```
cdata: This is temporary until our hashes settle.

cdata:
~~Current status: can't lint or build because the git dependency hashes are very broken. I think we will need to do some coordinated merging to fix it.~~

I was able to line up the SHAs just right to get it to keep building.

.cargo/config.toml
```diff
@@ -1,3 +1,3 @@
 [target.wasm32-unknown-unknown]
 runner = "wasm-bindgen-test-runner"
-rustflags = ['--cfg', 'getrandom_backend="wasm_js"']
+rustflags = ['--cfg', 'getrandom_backend="wasm_js"', '--cfg', 'web_sys_unstable_apis']
```
cdata, line 3: Needed to enable support for [Web Locks API](https://developer.mozilla.org/en-US/docs/Web/API/Lock)

Gozala: Can we add a comment for this in code so it's not lost.

rust/tonk-blobs/src/vfs/native.rs
```diff
@@ -0,0 +1,258 @@
...
+    /// The root path is used as-is; callers should ensure it is absolute and
+    /// already exists on disk before performing any I/O.
+    pub fn new(root: PathBuf) -> Self {
+        Self { root }
+    }
+
+    /// Converts a given path into a normalized path rooted on the internal `root`.
+    /// Prevents path traversal by resolving `..` components and verifying
+    /// the result stays within the root directory.
+    fn local_path(&self, given_path: &str) -> Result<PathBuf, TonkBlobsError> {
```
Gozala, line 42: Nit: not sure this is useful but I end up embracing file URIs in the FileSystem capability provider as a root to avoid having to deal with path resolution and OS level differences, which may not be a bad idea here either.

rust/tonk-blobs/src/vfs/native.rs
```diff
@@ -0,0 +1,258 @@
...
+        let joined = self.root.join(given_path.trim_start_matches('/'));
```
Gozala, line 43: Nit: `given_path.trim_start_matches('/')` this seems unintuitive to me, as in I would rather rather get error saying that give_path like `/passwords/tonk` is outside of authorized path as opposed to interpret it as relative `./password/tonk`.

rust/tonk-blobs/src/vfs/native.rs
Gozala, line 1: Nit: seems like this API does not handle races that are possible due to concurrency, which is probably by design but something perhaps docs should call out.

rust/tonk-blobs/src/vfs/web.rs
```diff
@@ -0,0 +1,536 @@
...
+            ));
+        }
+
+        let (dirs, file) = segments.split_at(segments.len() - 1);
+        Ok((dirs.to_vec(), file[0]))
+    }
+
+    /// Navigates from a starting directory through a chain of subdirectories,
+    /// optionally creating them along the way.
+    async fn navigate(
```
Gozala, line 215: Nit: I would find `resolve` to be more intuitive name for this.

rust/tonk-blobs/src/vfs/web.rs
```diff
@@ -0,0 +1,536 @@
...
+            })?
+            .dyn_into()
+            .map_err(|_| TonkBlobsError::Get("Expected File".into()))?;
+
+        Ok(FileReader::new(file.stream()))
+    }
+
+    /// Creates a file at the given path (relative to root) and returns a handle for writing.
+    /// Intermediate directories are created as needed.
+    pub async fn write_to(&self, path: &str) -> Result<FileWriter, TonkBlobsError> {
```
Gozala, line 304: Nit: I would fine `create_file_writer` or something along those lines more intuitive

rust/tonk-blobs/src/blob_storage.rs
```diff
@@ -0,0 +1,377 @@
...
+fn hash_to_path(hash: &Blake3Hash) -> String {
+    let b58 = hash.as_bytes().to_base58();
+    format!("{}/{}/{}/{}", &b58[0..2], &b58[2..4], &b58[4..6], b58)
+}
+
+/// High-level abstraction over platform-specific content-addressed blob
+/// storage.
+///
+/// Blobs are stored in a sharded directory tree derived from the base58
+/// encoding of their hash. Given a hash that encodes to `AbCdEfGh...`:
```
Gozala, line 19: Nit: It's not immediately obvious to me rational for sharing or this specific distribution, perhaps something worth capturing in here also.

rust/tonk-blobs/src/vfs.rs
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+    struct TestHarness {
+        vfs: Vfs,
+    }
+
+    #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+    fn harness() -> TestHarness {
+        let vfs = Vfs::new(format!("test/{}", ulid::Ulid::new()));
```
Gozala, line 43: How is `ulid` available here without it being imported anywhere ?

Gozala (approved):
Looks good to me. Suggested few updates to naming and some places where code comments may help.

valadaptive:
@cdata :wave: Hi! I noticed this PR when looking at https://github.com/whatwg/fs/pull/180 again.

You mentioned:

> On the web, `BlobStorage` uses the [origin-private file system](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system). This was mostly painless to use, except: OPFS does not have a standard atomic "move," (which the current implementation relies on to enable hashing without holding the whole file in memory). Thankfully, [Chrome has a non-standard `move` that the WG is attempting to standardize](https://github.com/whatwg/fs/pull/180). So on Chromium browsers, we have the fast move and on other browsers we have a fallback that implements a comparatively glacial read-and-pipe-to-writer scheme. It's probably still better than reading a very large file into memory, though.

The situation is actually better! All three browsers implement a "move file" operation in some form. They just implement *different subsets* of the interface:
- Firefox defines the `move(name)`, `move(parent)`, and `move(parent, name)` overloads on `FileSystemHandle` (the base class for both `FileSystemFileHandle` and `FileSystemDirectoryHandle`).
- Chromium defines the `move(name)`, `move(parent)`, and `move(parent, name)` overloads on `FileSystemFileHandle` only. You cannot move a directory.
- WebKit/Safari defines *only* the `move(parent, name)` overload, on `FileSystemHandle`.

So, you can actually use the `move` method in *all* browsers! You just need to stick to the subset that all browsers support currently: specifying both the new filename *and* parent directory, and only calling it on `FileSystemFileHandle`.

PR #399, opened 2026-02-12
feat(cli): TON-1863: bulk crud

jackddouglas:
### Description

- `tonk batch` subcommand for batch CRUD on instances of concepts in tonk cli
- each command has `--stdin`/`--file` flags for passing JSON data in via stdin or file respectively
- `batch create`/`batch update`/`batch delete`

### Related issues

- closes TON-1863

---

## PR-Codex overview
This PR introduces batch operations for managing instances in the `tonk` CLI, allowing users to create, update, and delete multiple instances of concepts in a single command.

### Detailed summary
- Added a new module `batch` for batch operations.
- Implemented `BatchCommands` enum with subcommands: `Create`, `Update`, `Delete`.
- Integrated batch operations into the main command flow.
- Created functions: `batch_create`, `batch_update`, `batch_delete` for respective operations.
- Updated documentation for batch commands and usage examples.

### DIFF
--- a/.claude/commands/tonk.md
+++ b/.claude/commands/tonk.md
@@ -96,6 +96,38 @@ tonk --json delete "did:key:z..."
 # → {"ok":true,"id":"did:key:z...","concept":"Task"}
 ```
 
+## Batch operations — `tonk batch`
+
+Create, update, or delete multiple instances atomically in a single commit. Input is a JSON array via `--file` or `--stdin`. If any item fails validation, the entire batch aborts with no changes committed.
+
+```bash
+# Batch create — array of attribute objects
+echo '[{"title":"Fix bug","status":"todo"},{"title":"Write docs","status":"todo"}]' \
+  | tonk --json batch create Task --stdin
+# → {"ok":true,"concept":"Task","count":2,"created":[{"id":"did:key:z...","data":{...}},...]}}
+
+# From a file
+tonk --json batch create Task --file tasks.json
+
+# Batch update — each object must include "id" plus fields to change
+echo '[{"id":"did:key:z...","status":"done"},{"id":"did:key:z...","priority":"high"}]' \
+  | tonk --json batch update Task --stdin
+# → {"ok":true,"concept":"Task","count":2,"updated":[{"id":"did:key:z...","updated":[{"status":"done"}]},...]}}
+
+# Batch delete — array of instance ID strings
+echo '["did:key:z...","did:key:z..."]' \
+  | tonk --json batch delete Task --stdin
+# → {"ok":true,"concept":"Task","count":2,"deleted":["did:key:z...","did:key:z..."]}
+```
+
+### Input formats
+
+| Operation | Input shape | Required fields |
+|-----------|------------|-----------------|
+| `batch create` | `[{...}, ...]` | Attribute key-value objects |
+| `batch update` | `[{...}, ...]` | Each object must have `"id"` plus fields to change |
+| `batch delete` | `["id", ...]` | Array of instance ID strings |
+
 ## Spaces
 
 ```bash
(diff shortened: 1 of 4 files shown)


cdata (commented):
What is the test plan for these commands?

jackddouglas:
@cdata i'm working on a follow up pr that has integration tests for each path through the cli. for login, we'll need a helper that bypasses the browser passkey flow, but everything else should be fine

jackddouglas:
see  [feat(cli): TON-1867: integration tests](https://github.com/tonk-labs/tonk/pull/401)

.claude/commands/tonk.md
```diff
@@ -96,6 +96,38 @@ tonk --json delete "did:key:z..."
 # → {"ok":true,"id":"did:key:z...","concept":"Task"}
 ```
 
+## Batch operations — `tonk batch`
+
+Create, update, or delete multiple instances atomically in a single commit. Input is a JSON array via `--file` or `--stdin`. If any item fails validation, the entire batch aborts with no changes committed.
```
Gozala, line 101: Nit: My intuition is hat you don't want to abort but rather ask clarifying questions as that reduces context use. E.g. we could always ingest whatever json even if it does not match any concepts and then point out ones that don't match things and ask what to do with them.

This enables second tool use to only send corrected things as opposed to redo the whole thing

.claude/commands/tonk.md
```diff
@@ -96,6 +96,38 @@ tonk --json delete "did:key:z..."
...
+
+```bash
+# Batch create — array of attribute objects
+echo '[{"title":"Fix bug","status":"todo"},{"title":"Write docs","status":"todo"}]' \
+  | tonk --json batch create Task --stdin
```
Gozala, line 106: Nit: I personally would find `import` as opposed to `task` to be more intuitive name

.claude/commands/tonk.md
```diff
@@ -96,6 +96,38 @@ tonk --json delete "did:key:z..."
...
+# → {"ok":true,"concept":"Task","count":2,"deleted":["did:key:z...","did:key:z..."]}
+```
+
+### Input formats
+
+| Operation | Input shape | Required fields |
+|-----------|------------|-----------------|
+| `batch create` | `[{...}, ...]` | Attribute key-value objects |
+| `batch update` | `[{...}, ...]` | Each object must have `"id"` plus fields to change |
+| `batch delete` | `["id", ...]` | Array of instance ID strings |
```
Gozala, line 129: Nit: Could we use terms assert / retract instead of update delete ?

rust/tonk-cli/src/batch.rs
```diff
@@ -0,0 +1,464 @@
...
+                        .iter()
+                        .map(|a| short_attribute(&stored_name, a))
+                        .collect::<Vec<_>>()
+                        .join(", ")
+                );
+            }
+            qualified_fields.push((qualified, value_str));
+        }
+
+        let instance_entity = Entity::new().context("Failed to generate instance entity")?;
```
Gozala, line 88: I would strongly advise against generating identifiers it's unnecessary source of entropy. It would make a lot more sense to take derive unique identifier from the hash. This way ingesting same data will produce same identifier as opposed to new one each time around.

But also we should probably just have a way to decode JSON into concept instances

rust/tonk-cli/src/batch.rs
```diff
@@ -0,0 +1,464 @@
...
+            }));
+        }
+
+        // Back-reference from concept to instance
+        instructions.push(Instruction::Assert(Artifact {
+            the: Attribute::from_str(ATTR_CONCEPT_INSTANCE)?,
+            of: concept.clone(),
+            is: Value::Entity(instance_entity.clone()),
+            cause: None,
+        }));
```
Gozala, line 122: I would recommend not going such a low level route and instead embrace `Relation` struct as used shown here https://github.com/dialog-db/dialog-db/blob/tonk/rust/dialog-query/src/session/transaction.rs#L262-L277

There is also [.commit(transaction)](https://github.com/dialog-db/dialog-db/blob/tonk/rust/dialog-query/src/session.rs#L166) that can be used to commit

rust/tonk-cli/src/batch.rs
```diff
@@ -0,0 +1,464 @@
...
+                }));
+            }
+
+            // Assert new value
+            instructions.push(Instruction::Assert(Artifact {
+                the: Attribute::from_str(&qualified)?,
+                of: entity.clone(),
+                is: new_value,
+                cause: None,
+            }));
```
Gozala, line 266: If you want to replace fact you can use causal link to it so that new value would succeed it. There is also nuance of cardinality as in we don't want to replace if cardinality is many.

Gozala (approved):
Can you please include some tests in the change so we can ensure that things are working and we're not regress it without realizing

.claude/commands/tonk.md
jackddouglas: Task is the name of the concept, though this has since been changed since concepts are duck-typed

rust/tonk-cli/src/batch.rs
jackddouglas: addressed in refactor

PR #403, opened 2026-02-20
refactor(cli)

jackddouglas:
## Summary
The thrust of this refactor is to bring `tonk-cli` closer to `dialog-db`'s conceptual model, primary driven by the need to derive identity from structure, not randomness. Fact entity IDs are now deterministically derived from field content, including attribute and value. Attributes are sorted by name, blake3 hashed, then an Ed25519 key is derived and encoded into a `did:key`. Concept entities are derived from their structure, including attributes. This way the same data converges to the same identity.

The central registry was also removed in favor of structural discovery. This eliminates the need to search for concepts by name via `registry/concept` or rules via `registry/rule` etc. Similarly, entity-to-concept membership is no longer tracked with explicit back-references. Instead, an entity belongs to a concept if it has values for all of that concept's attributes.

When defining a new concept, three outcomes are possible:
1. If no existing concept with this structure/name: create concept entity and commit
2. If same name and attributes: convergent define, the description is re-asserted
3. If same name, different attributes: in interactive mode the user is prompted to update or cancel, in JSON a structured conflict response is returned

## Other changes
Terminology was also updated:
- `update` -> `assert`
- `delete` -> `retract`
- `space set` -> `space load`

---

## PR-Codex overview
This PR focuses on renaming concepts from `instances` to `entities`, updating related documentation and function signatures, and improving the handling of spaces and batch operations in the `tonk` CLI tool.

### Detailed summary
- Added `pub mod entity;` to `lib.rs`.
- Renamed `Keystore::new()` documentation from "keystore instance" to "keystore entity".
- Removed references to `SafeMeal`, replacing it with `Meal` in YAML files.
- Updated `rust/tonk-cli/src/instance.rs` to simplify session registration.
- Simplified query answer handling by removing entity ID deduplication.
- Refactored import functions to pass `SpaceContext`.
- Updated `space.rs` to change `set` to `load` for active space selection.
- Introduced `open` method to load or create spaces.
- Renamed `instance` references to `entity` across various files.
- Adjusted command descriptions in the CLI for clarity.
- Improved batch operation handling for entities in `batch.rs`.

> The following files were skipped due to too many changes: `rust/tonk-cli/src/batch.rs`, `rust/tonk-cli/src/rule.rs`, `rust/tonk-cli/src/entity.rs`, `rust/tonk-cli/src/import/commit.rs`, `rust/tonk-cli/src/schema.rs`, `rust/tonk-cli/src/concept.rs`, `rust/tonk-cli/tests/cli_integration.rs`


.claude/commands/tonk.md
```diff
@@ -1,5 +1,5 @@
 # tonk CLI — Agent Reference
-tonk is a CLI for managing spaces (context repositories) with DID-based identity. Spaces store structured data through **concepts** (schemas) and **instances** (entries). All commands support `--json` for structured output.
+tonk is a CLI for managing spaces (context repositories) with DID-based identity. Spaces store structured data through **concepts** (schemas) and **entities** (entries). All commands support `--json` for structured output.
```
Gozala, line 2: I think this violates my (admittedly new) [mental model](https://gozala.io/dialog/anatomy) where


- **Statement** is what you assert or retract, it could be a _conclusion_ you've reached or a primitive _association_.
- **Claim** is what the system captures from your _statement_ and attaches a local (logical) timestamp, it's an active assumption / belief.
- **Conclusion** is what gets materialized when you query for a _Concept_, the system collects evidence from stored _claims_ and assembles it according to the conceptual model.

Association is the `{the,of,is}` that relates some value with entity through nominal identifier `the`.

.claude/commands/tonk.md
```diff
@@ -20,37 +20,59 @@ Always run `tonk --json status` first to orient before creating new sessions/spa
 
 ## Concepts — defining schemas
 
-A concept is a named schema that defines what attributes instances can have. Attribute names are auto-prefixed with the concept's lowercase namespace (e.g. concept "Task" → attributes stored as `task/title`, `task/status`).
+A concept is a named schema that defines what attributes entities can have. Attribute names are auto-prefixed with the space name (for CLI-defined concepts) or the YAML-declared namespace (for imported concepts). For example, in a space named "my-space", concept "Task" → attributes stored as `my-space/title`, `my-space/status`. Description is required (`--description` flag).
```
Gozala: Might be worth reviewing this thread https://discord.com/channels/1423258342578589749/1476271563174183023, as it offers some justification for concrete terminology.

I have a feeling that using a space name for prefixing is going to be a mistake, it would introduce divergence and make it difficult to have cross space queries making every concept unique to a space.

.claude/commands/tonk.md
```diff
@@ -20,37 +20,59 @@ Always run `tonk --json status` first to orient before creating new sessions/spa
...
 
 ```bash
 # List all concepts in the active space
 tonk --json concept
-# → [{"name":"Task","instances":12},{"name":"Contact","instances":5,"description":"People and orgs"}]
+# → [{"name":"Task","entities":12},{"name":"Contact","entities":5,"description":"People and orgs"}]
```
Gozala: I think `count` instead of `entries` would be more neutral and clear. Alternatively I'd say `matches` better captures semantics.

.claude/commands/tonk.md
Gozala: Also worth calling out this seems to contradict examples below that show `task/status` etc...

.claude/commands/tonk.md
```diff
@@ -20,37 +20,59 @@ Always run `tonk --json status` first to orient before creating new sessions/spa
...
-# → {"ok":true,"name":"Task","attributes":["task/title","task/status","task/priority"]}
-
-# With a description
-tonk --json concept define Contact name email role --description "People and organizations"
+tonk --json concept define Task title status priority --description "A task to track"
+# → {"ok":true,"name":"Task","attributes":["task/title","task/status","task/priority"],"description":"A task to track"}
 
 # Interactive mode (prompts for attributes one by one)
-tonk concept define Task
+tonk concept define Task --description "A task to track"
```
Gozala, line 35: What do you think of `tonk assert concept Task` instead ? For what it's worth I'm hoping that in dialog you'd be able to do `tx.assert(my_concept)` directly.

I also like assert better because it is not necessarily defining new concept it could also be updating existing ones.

Gozala, line 35: What do you think of `tonk assert concept Task` instead ? For what it's worth I'm hoping that in dialog you'd be able to do `tx.assert(my_concept)` directly.

I also like assert better because it is not necessarily defining new concept it could also be updating existing ones.

Gozala, line 35: What do you think of `tonk assert concept Task` instead ? For what it's worth I'm hoping that in dialog you'd be able to do `tx.assert(my_concept)` directly.

I also like assert better because it is not necessarily defining new concept it could also be updating existing ones.

.claude/commands/tonk.md
```diff
@@ -20,37 +20,59 @@ Always run `tonk --json status` first to orient before creating new sessions/spa
...
 
 # Show concept schema
 tonk --json concept show Task
-# → {"name":"Task","attributes":["task/title","task/status","task/priority"],"instance_count":12,"entity":"did:key:z..."}
+# → {"name":"Task","attributes":["task/title","task/status","task/priority"],"entity_count":12,"entity":"did:key:z..."}
```
Gozala: I think show is misleading here. I would expect concept description in formal JSON notation. Match count seems more like `concept analyze Task` that can provide various stats about it.

.claude/commands/tonk.md
```diff
@@ -20,37 +20,59 @@ Always run `tonk --json status` first to orient before creating new sessions/spa
...
+tonk --json concept define Task title status priority due_date --description "Updated task schema"
+# → {"ok":false,"conflict":true,"name":"Task",
+#    "existing_entity":"did:key:z...",
+#    "existing_attributes":["title","status","priority"],
+#    "proposed_entity":"did:key:z...",
+#    "proposed_attributes":["title","status","priority","due_date"],
+#    "message":"A different concept already exists under the name 'Task'. Re-run with --update to replace it (a provenance link will be created), or choose a different name."}
+```
+
+When a conflict is detected in JSON mode, the caller should inspect `existing_attributes` vs `proposed_attributes` and decide whether to update or rename. In interactive mode (without `--json`), the user is prompted to choose and can provide a rationale for the update, which is stored as `concept/update-rationale` on the new concept entity. Updated concepts link to their predecessor via `concept/prior` for provenance tracking.
```
Gozala, line 73: This does not feel like a great DX perhaps side effect of define conflating defining a concept and publishing it under certain name.

I wonder if we'd be better of having something like

`tonk --json concept assert <JSON>` 

I would expect input in a formal JSON notation.

Then you could have an optional param `--publish <name>` to publish it under specific name

`tonk --json concept assert <JSON> --publish <name>`

This way concept is still asserted, but if there is already one published under that name it can still

1. Assert concept
2. Produce error showing existing concept under the same name.
3. Ask if you want to update, choose different name or not name it at all.

At the same time you could then have `--update` flag to indicate awareness that it replaces existing concept

`tonk --json concept assert <JSON> --publish <name> --update`

P.S. I would recommend not erroring as best practice and instead ask clarifying questions. That helps LLM not to go in retry loops inflating tool calls and context use. It's far less overhead to react to name already exists want to call it something else than trying to look at a diff and and come up with a alternative

.claude/commands/tonk.md
```diff
@@ -64,12 +86,12 @@ echo '{"title":"Fix login bug","status":"todo"}' | tonk --json create Task --std
...
 
-The instance ID (`did:key:z...`) is randomly generated and returned in the response. Store it to reference the instance later.
+The entity ID (`did:key:z...`) is deterministically derived from the entity's field content and returned in the response. Store it to reference the entity later.
 
-## Querying instances — `tonk query`
+## Querying entities — `tonk query`
 
 ```bash
-# All instances of a concept
+# All entities of a concept
```
Gozala, line 94: Seems like this should be `concept query` as we may offer `claim query` also ?

.claude/commands/tonk.md
```diff
@@ -80,25 +102,27 @@ tonk --json query Task status=todo
 tonk --json query Task status=todo priority=high
 ```
 
-## Instance operations — `tonk show`, `tonk update`, `tonk delete`
+## Entity operations — `tonk show`, `tonk assert`, `tonk retract`
 
 ```bash
-# Show full details of an instance
+# Show full details of an entity
 tonk --json show "did:key:z..."
```
Gozala, line 109: I'm inlined so say that perhaps this also should be `analyze`.

What happens when

1. When several concepts match this entity ?
2. When there are attributes that aren't part of any concept ?

I'm inclined to say that analyze should provide stats for all of them like what are concepts does entity match and all the attributes associated with an entity

.claude/commands/tonk.md
```diff
@@ -80,25 +102,27 @@ tonk --json query Task status=todo
...
 # → {"id":"did:key:z...","concept":"Task","data":{"title":"Fix login bug","status":"todo","priority":"high"},"created":1739012345}
 
-# Update specific fields (keys auto-prefixed)
-tonk --json update "did:key:z..." status=done
+# Assert (update) specific fields (keys auto-prefixed)
+tonk --json assert "did:key:z..." status=done
```
Gozala, line 113: I think here also it's either `tonk concept assert Task  did:key:z... status=done` because there might be multiple concepts associated with an entity.

Similarly we could have then lower level interface for asserting attributes directly that is perhaps

`tonk attribute assert --the task/status --of  did:key:z... --is done` 

Or if you want to stick to similar notation 

`tonk attribute assert task did:key:z... status=done`

P.S.: It would be worth benchmarking if similarity would inflict **proactive interference** or not.

Gozala: Alternative could be notation like

```
tonk assert --attribute todo --of did:key:z... status=done
tonk assert --concept Task --of did:key:z... status=done
```

First one translating to `todo/status=done` and later resolving concept `Task`'s `status` and asserting that attribute.

.claude/commands/tonk.md
```diff
@@ -114,7 +138,7 @@ echo '[{"id":"did:key:z...","status":"done"},{"id":"did:key:z...","priority":"hi
   | tonk --json batch update Task --stdin
```
Gozala, line 138: I think `tonk import --concept Task --stdin --json` seems a lot more intuitive for me. You could than also have flags to tell if it should update existing assertion or only import new ones.

Gozala: That way you could also have `--retract` flag implying that you're importing retractions. Better yet you could actually have subset of fields there you wish to retract as opposed to retract everything.

I would also suggest instead of array you could instead key by entity instead.

.claude/commands/tonk.md
```diff
@@ -137,26 +161,26 @@ tonk --json space create "name"      # Create a new space
...
+## Rules — deriving entities from patterns
 
-A rule defines how instances of a concept can be **derived** from patterns across existing facts. Rules use Datalog-style deduction: positive premises (`when`) that must hold, and negative premises (`unless`) that must NOT hold. Variables (prefixed with `?`) bind values and create implicit joins across premises.
+A rule defines how entities of a concept can be **derived** from patterns across existing facts. Rules use Datalog-style deduction: positive premises (`when`) that must hold, and negative premises (`unless`) that must NOT hold. Variables (prefixed with `?`) bind values and create implicit joins across premises.
 
-Multiple rules can derive the same concept — they act as **OR branches**. Each rule independently produces derived instances, and results are merged. This lets you evolve rules independently rather than maintaining one giant rule with many branches.
+Multiple rules can derive the same concept — they act as **OR branches**. Each rule independently produces derived entities, and results are merged. This lets you evolve rules independently rather than maintaining one giant rule with many branches.
```
Gozala: I think you overload entity here. Entities are identifiers what rules do is derive conclusions that adhere to establish conceptual model. Hence multiple rules describing how to derive conclusions naturally yield more conclusions

.claude/commands/tonk.md
```diff
@@ -137,26 +161,26 @@ tonk --json space create "name"      # Create a new space
...
 
-When you query a concept that has rules, derived instances are transparently merged with stored instances.
+When you query a concept that has rules, derived entities are transparently merged with stored entities.
```
Gozala: I'd say when you query a concept deduced by multiple rules all of their conclusions are combined forming unified set.

rust/tonk-cli/examples/rules.yaml
```diff
@@ -20,15 +20,15 @@ user.rules:
...
         attendee: ?person
         recipe: ?recipe
         occasion: ?event
     when:
       - diy.planner/Event:
           this: ?event
           title: ?title
-      - diy.planner/SafeMeal:
+      - diy.planner/Meal:
           this: ?safe
```
Gozala: ⚠️ This looks like recursive rule without negation. Recursive rules are not currently supported by dialog, but even if they were this would be a kind of rule that dialog would reject at compile time even if technically it is a valid datalog (JS implemented query engine does actually reject such deductive rules). That is because this rule produces nothing new and therefor is unlikely to represent authors intent. More concretely it deduces subset of Meals, but then union of the set with it's subset is still the same set, meaning it derives nothing new.

rust/tonk-cli/src/entity.rs
Gozala, line 1: Calling this entity is really misleading, it allows you to query / and update concept matches a.k.a conclusions. I think it's more closer to what you get in SQL when you run a query with a JOIN(s). You won't call it row because it's not from one table. Closest thing is a view but that term is so overloaded that I would definitely not go there.

I think calling result of querying a concept can be best described as either **match**, that is entities that match the conceptual model along with supporting evidence. Or we can call it a "conclusion", as in you query knowledge substrate for conclusions matching queried conceptual model.

Gozala (changes requested):
I skimmed it through. Only rejecting because I think rule examples are misleading and need to be addressed. Feel free to merge without another review after addressing this.

I also provided feedback on some terminology and a CLI interface, but feel free to disregard or address in a followup PRs.

PR #405, opened 2026-02-24
fix(cli): flatten and require `this` binding, rule names optional

jackddouglas:
## Summary
`this` field in YAML rules was previously provided to `tonk-cli` outside of other bindings, meaning `this` could appear in two different places depending on the context: either passed separately in parsed YAML or in bindings. If omitted, `build_rename_map` would auto-detect by scanning for the first unclaimed entity variable which was confusing, fragile, and could lead to subtle bugs where the auto-detected entity variable wasn't what the user intended.

For instance, if the following `when` premises was passed to the rules system:
```yaml
when:
  - diy.cook/Recipe:
      this: ?recipe
      ingredient: ?ingredient
  - diy.health/Allergy:
      this: ?allergy
      person: ?person
```
auto-detection would pick `?recipe` as the entity variable when `?allergy` may have made more semantic sense.

By requiring `this` in rule bindings, the deduced entity is always explicit.

## Other Changes
Rule names are now optional, and entities are deterministically derived from `blake3(definition_json)` ensuring all rules are content-addressed and idempotent.

---

## PR-Codex overview
This PR primarily focuses on improving the handling of rules in the `tonk-cli` application by clarifying the usage of the `this` variable in rule definitions and enhancing the command-line interface for rule management.

### Detailed summary
- Removed `this` from conclusion bindings in rules.
- Updated rules to use `?this` directly in premises.
- Enhanced commands to accept entity IDs for rule management.
- Improved error messages for rule definition and validation.
- Updated documentation and comments for clarity.


rust/tonk-cli/src/rule.rs
```diff
@@ -59,31 +59,35 @@ impl RuleDefinition {
...
         }
 
+        if !def.conclusion.bindings.contains_key("this") {
+            anyhow::bail!(
+                "Rule conclusion bindings must include a 'this' key that specifies \
+                 which entity variable provides the derived entity's identity. \
+                 For example: \"bindings\": {{ \"this\": \"?entity\", ... }}"
+            );
+        }
+
```
Gozala, line 71: Just flagging that this is at odds with https://github.com/dialog-db/dialog-db/pull/208/changes#diff-356d33e1251033883caf1002e96a6bf674e0ee948b46f74001f2fc715538808fR406 which says that `?this` is implicitly bound to entity.

Gozala (approved):
I have skimmed through this. Overall I think it's improvement over previous behavior, yet I would very much like to align with https://github.com/dialog-db/dialog-db/pull/208 as I think it offers more thought out interface. That said I think we should land this and consider alignment afterwards.

PR #422, opened 2026-04-14
feat: rewrite tonk-worker to use dialog-repository

Gozala:
Replace tonk-space with dialog-repository/dialog-operator.

Deleted modules (replaced by dialog built-ins):
- storage.rs (custom IndexedDB backend -> dialog Storage)
- key_store.rs (Ed25519 key persistence -> dialog Profile)
- identity.rs (Identity wrapper -> dialog Profile)
- session.rs (Space + Operator bundle -> Profile + Operator)
- account.rs (known spaces metadata -> out of scope for now)

Rewritten modules:
- worker.rs: TonkState now holds Profile + Operator<DefaultSpace>
  initialized via Profile::open/derive/build pattern
- router.rs: New URL structure /api/repository/{repo}/branch/{branch}/...
- claim.rs (was fact.rs): Uses branch.transaction() for writes, branch.claims().select() for reads
- init.rs (was authorize.rs): Sets up UCAN remote + upstream
- sync.rs: Uses branch.pull()/push().perform(&operator)
- status.rs: Reports repo DID and upstream state
- identify.rs: Returns profile DID
- inspect/branch.rs: Branch revision details
- inspect/remote.rs (was inspect/site.rs): Remote address + branch info

Dropped routes:
- /api/delegations (delegation management now handled by dialog)
- Archive block fetch endpoint (low-level debugging)

### Description

_A few sentences describing the overall effects and goals of the pull request's commits.
What is the current behavior, and what is the updated/expected behavior with this PR?_

### Other changes

### Tested

### Related issues

### Backwards compatibility

### Documentation

---

## PR-Codex overview
This PR focuses on refactoring the `tonk-worker` codebase by removing unused modules, updating API routes, and enhancing error handling. It also introduces new endpoints for managing repositories and branches, streamlining the synchronization process.

### Detailed summary
- Deleted multiple unused files related to `account`, `session`, and `storage`.
- Updated `Cargo.toml` to remove outdated dependencies and add new ones such as `dialog-capability` and `dialog-effects`.
- Refactored the `identify` endpoint to return user profile DID instead of user DID.
- Introduced new `init` endpoint to set up UCAN remote and upstream for branches.
- Modified `sync` operations to include repository and branch parameters in API routes.
- Enhanced error handling in the `TonkWorkerError` enum with a new `NotFound` variant.
- Updated the `StatusResponse` to include repository name and operator DID.
- Added new inspection routes for branches, remotes, and archives.
- Updated tests to reflect new API routes and functionalities.

> The following files were skipped due to too many changes: `rust/tonk-worker/src/router/claim.rs`


rust/tonk-worker/src/lib.rs
```diff
@@ -24,91 +24,47 @@
...
+    const origSet = desc.set;
+    Object.defineProperty(IDBDatabase.prototype, 'onversionchange', {
+        set(handler) {
+            origSet.call(this, function() { this.close(); });
+        },
+        get: desc.get,
+        configurable: true,
+    });
+}
+"#)]
```
Gozala, line 49: @cdata can you please offer some advice here. This feels like horrible hack, yet without something like this I keep running into a problem with `onversionchange` producing errors about dropped closure, which happens on reload. From what I can gather it's a service worker that is being replaced is dropped but even listener seems to stick around. Furthermore error in that event handler seems to prevent db from opening in the new worker because all prior connections need to close.

This hack addresses the problem, but I hate it and would love to find a better solution.

P.S. It did not occur before because previously we had fixed set of stores and version change was never triggered. In a new design stores are populated on demand, which trigger version change when new stores are opened.

cdata: Is it a problem because we don't set the handler (`idb` does, and therefor we can't remove it at the intended time)?

If that's the case, I would file an issue against the `idb` crate, and also a tracking issue here, and then just keep the hack in place until we can clean the handler up "properly."

Gozala:
Superseded by #428

PR #426, opened 2026-04-17
feat(invite): shared invite URL and claim orchestration

jackddouglas:
- new `tonk-invite` crate provides invite URL format and claim orchestration for native and browser
- both `carry` and `tonk-ui` will depend on it so invite URLs issued by one are redeemable by the other, making them wire-compatible
- requires that invites target a specific repo, but `InviteAudience::{Open, Scoped}` allows for open-ended invites if no user DID provided, and scoped invites if one is

---

## PR-Codex overview
This PR introduces the `tonk-invite` crate, which facilitates the creation and management of invite URLs for a service. It adds relevant dependencies and implements core functionality for generating and parsing invite URLs, including audience management.

### Detailed summary
- Added `tonk-invite` to `Cargo.toml` and workspace dependencies.
- Created `rust/tonk-invite/Cargo.toml` with metadata and dependencies.
- Implemented `Invite` struct in `rust/tonk-invite/src/lib.rs` for managing invites.
- Added methods for creating, parsing, and claiming invites.
- Defined `InviteAudience` enum for audience management (open/scoped).
- Introduced error handling for invite URL parsing and validation.
- Implemented tests for invite functionality, including URL round-tripping and audience checks.

> The following files were skipped due to too many changes: `Cargo.lock`

### DIFF
--- a/rust/tonk-invite/src/lib.rs
+++ b/rust/tonk-invite/src/lib.rs
@@ -286,7 +286,7 @@ impl Invite {
                 let chain_audience = self.chain.audience();
                 anyhow::ensure!(
                     *chain_audience == *audience,
-                    "this scoped invite was issued to {}, but the redeemer is {}; \
+                    "this invite is for {} and cannot be redeemed by {}; \
                      ask the inviter to issue a new invite for your DID, \
                      or switch to the identity the invite was issued to",
                     chain_audience,
@@ -312,7 +312,16 @@ impl From<Invite> for UcanProof {
 }
 
 /// Outcome of [`Invite::claim`]: a delegation chain ready to be persisted
-/// by the caller, plus metadata carried over from the invite.
+/// by the caller, paired with the invite's sync remote.
+///
+/// We return this wrapper rather than a bare [`DelegationChain`] so that
+/// the `Subject::Specific` invariant validated at [`Invite`] construction
+/// carries through to [`ClaimedInvite::subject`] without re-checking, and
+/// so `remote_url` threads through to callers that want to persist sync
+/// config alongside the chain.
+///
+/// `#[non_exhaustive]` reserves room to grow additional carry-over fields
+/// (e.g. capability metadata) without a breaking change.
 #[derive(Debug)]
 #[non_exhaustive]
 pub struct ClaimedInvite {
@@ -341,7 +350,7 @@ mod tests {
     use dialog_varsig::principal::Principal;
 
     #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
-    use wasm_bindgen_test::{wasm_bindgen_test, wasm_bindgen_test_configure};
+    use wasm_bindgen_test::wasm_bindgen_test_configure;
 
     #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
     wasm_bindgen_test_configure!(run_in_browser);
@@ -374,33 +383,26 @@ mod tests {
         Ed25519Signer::import(seed).await.unwrap()
     }
 
-    #[cfg_attr(not(all(target_arch = "wasm32", target_os = "unknown")), test)]
-    #[cfg_attr(all(target_arch = "wasm32", target_os = "unknown"), wasm_bindgen_test)]
-    fn parse_rejects_non_url_input() {
+    #[dialog_common::test]
+    fn it_rejects_non_url_input() {
         let err = Invite::parse_url("not a url").unwrap_err();
         assert!(err.to_string().contains("not a valid URL"), "{err}");
     }
 
-    #[cfg_attr(not(all(target_arch = "wasm32", target_os = "unknown")), test)]
-    #[cfg_attr(all(target_arch = "wasm32", target_os = "unknown"), wasm_bindgen_test)]
-    fn parse_rejects_missing_access_parameter() {
+    #[dialog_common::test]
+    fn it_rejects_missing_access_parameter() {
         let err = Invite::parse_url("https://tonk.xyz/join").unwrap_err();
         assert!(err.to_string().contains("`access`"), "{err}");
     }
 
-    #[cfg_attr(not(all(target_arch = "wasm32", target_os = "unknown")), test)]
-    #[cfg_attr(all(target_arch = "wasm32", target_os = "unknown"), wasm_bindgen_test)]
-    fn parse_rejects_invalid_base58_in_access() {
+    #[dialog_common::test]
+    fn it_rejects_invalid_base58_in_access() {
         let err = Invite::parse_url("https://tonk.xyz/join?access=!!!not-b58!!!").unwrap_err();
         assert!(err.to_string().contains("valid base58"), "{err}");
     }
 
-    #[cfg_attr(
-        not(all(target_arch = "wasm32", target_os = "unknown")),
-        tokio::test(flavor = "current_thread")
-    )]
-    #[cfg_attr(all(target_arch = "wasm32", target_os = "unknown"), wasm_bindgen_test)]
-    async fn parse_rejects_wrong_length_fragment() {
+    #[dialog_common::test]
+    async fn it_rejects_wrong_length_fragment() {
         let subject = signer(&SUBJECT_SEED).await.did();
         let audience = signer(&AUDIENCE_SEED).await.did();
         let chain = make_chain(&ISSUER_SEED, &audience, &subject).await;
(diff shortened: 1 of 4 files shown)


rust/tonk-invite/src/lib.rs
```diff
@@ -0,0 +1,476 @@
...
+
+/// Outcome of [`claim`]: a delegation chain ready to be persisted by the
+/// caller, plus metadata about the invite.
+#[derive(Debug)]
+#[non_exhaustive]
+pub struct ClaimedInvite {
+    /// Final delegation chain terminating at the redeemer's DID.
+    pub chain: DelegationChain,
+    /// Repo subject DID (always `Subject::Specific`).
+    pub subject: Did,
```
Gozala, line 88: This seems redundant as chain subject will have this no ?

rust/tonk-invite/src/lib.rs
```diff
@@ -0,0 +1,476 @@
...
+    let chain = DelegationChain::try_from(chain_bytes.as_slice()).with_context(|| {
+        format!(
+            "invite `access` parameter did not decode to a valid delegation chain ({} bytes)",
+            chain_bytes.len()
+        )
+    })?;
+    let subject = chain.subject().cloned().context(
+        "invite delegation chain must target a specific repo subject; \
+         chains with Subject::Any are not valid invites",
+    )?;
```
Gozala, line 138: If no subject is present you could treat issuer of the root delegation as subject because they implicitly delegate access to their own did.

rust/tonk-invite/src/lib.rs
Gozala, line 1: I think it would be nice to have something like `Invite` struct which could then have methods to format it as URL. In the future we could also then add maybe serialize as invite file or QR code etc.

P.S.: I think `base` url should not be part of the `Invite` instead it's something you provide when you're serializing invite into URL.

rust/tonk-invite/src/lib.rs
```diff
@@ -0,0 +1,476 @@
...
+        .subject(UcanSubject::Specific(subject))
+        .command(vec![])
+        .try_build()
+        .await
+        .context("failed to build redelegation")?;
+
+    chain
+        .push(delegation)
+        .context("failed to extend delegation chain with redelegation")
+}
```
Gozala, line 253: Would we nice if we could maybe go from invite to [`UcanProof`](https://github.com/dialog-db/dialog-db/blob/ae204eaea39dd960ec457715640d5c59fb8d40d5/rust/dialog-ucan/src/access.rs#L86-L116) because that would allow you to have nice API for, claiming, redelegation and saving it into your profile as shown here

https://github.com/dialog-db/dialog-db/tree/feat/query-integration/rust/dialog-ucan

Gozala: E.g if we had that `Invite` struct we could then have `impl From<Invite> for UcanProof` and you'd get .claim on it without having to reimplement it.

Gozala: I would expect than Invite may have method that returns optional Signer

rust/tonk-invite/src/lib.rs
```diff
@@ -0,0 +1,476 @@
...
+    fn parse_rejects_invalid_base58_in_access() {
+        let err = parse_invite_url("https://tonk.xyz/join?access=!!!not-b58!!!").unwrap_err();
+        assert!(err.to_string().contains("valid base58"), "{err}");
+    }
+
+    #[cfg_attr(
+        not(all(target_arch = "wasm32", target_os = "unknown")),
+        tokio::test(flavor = "current_thread")
+    )]
+    #[cfg_attr(all(target_arch = "wasm32", target_os = "unknown"), wasm_bindgen_test)]
```
Gozala, line 358: Why opt out of wasm ?

rust/tonk-invite/src/lib.rs
```diff
@@ -0,0 +1,476 @@
...
+            msg.contains("exactly 32 bytes") || msg.contains("got 3"),
+            "{msg}"
+        );
+    }
+
+    #[cfg_attr(
+        not(all(target_arch = "wasm32", target_os = "unknown")),
+        tokio::test(flavor = "current_thread")
+    )]
+    #[cfg_attr(all(target_arch = "wasm32", target_os = "unknown"), wasm_bindgen_test)]
```
Gozala, line 381: Why opt out of wasm ?

Gozala (changes requested):
Provided some suggestions, but those could be future improvements. Not approving only because bunch of tests opt-out of wasm which I think is a mistake

rust/tonk-invite/src/lib.rs
jackddouglas: it's derivable but wrapped in `Option` and since `subject` should be invariant i'd rather make that explicit in the type (topic of subject invariance up for debate per your comment below though)

rust/tonk-invite/src/lib.rs
jackddouglas: definitely if we assume the delegation chains only live within tonk assumptions, but seems wrong to impose those assumptions on dialog's `DelegationChain` when that's not necessarily the case? since inviter should always know what repo they're creating an invite for, it costs nothing to ensure we're not ambiguously reinterpreting keys and require an explicit subject, no?

rust/tonk-invite/src/lib.rs
```diff
@@ -0,0 +1,564 @@
...
+                    .delegate(audience.clone())
+                    .await
+                    .map_err(|e| anyhow::anyhow!("failed to redelegate invite: {e}"))?;
+                delegation.into_chain()
+            }
+            None => {
+                let chain_audience = self.chain.audience();
+                anyhow::ensure!(
+                    *chain_audience == *audience,
+                    "this scoped invite was issued to {}, but the redeemer is {}; \
```
Gozala: I think language is bit cryptic, I'd say something along the lines of "this invite is for {chain_audience} and can not be redeemed by {audience}".

rust/tonk-invite/src/lib.rs
```diff
@@ -0,0 +1,564 @@
...
+    fn from(invite: Invite) -> Self {
+        let scope = Scope::from_chain(&invite.chain);
+        UcanProof::from_chain(&invite.chain, scope)
+    }
+}
+
+/// Outcome of [`Invite::claim`]: a delegation chain ready to be persisted
+/// by the caller, plus metadata carried over from the invite.
+#[derive(Debug)]
+#[non_exhaustive]
```
Gozala, line 326: Can you add maybe add code comment on this, I'm not sure I understand intention here.

rust/tonk-invite/src/lib.rs
```diff
@@ -0,0 +1,564 @@
...
+    /// # Errors
+    ///
+    /// Returns an error if the chain's subject is not
+    /// [`Subject::Specific`][dialog_ucan_core::subject::Subject::Specific]
+    /// — tonk invites must always target a specific repo.
+    pub fn new(
+        chain: DelegationChain,
+        audience: InviteAudience,
+        remote_url: Option<Url>,
+    ) -> Result<Self> {
```
Gozala, line 99: Can we please call this something else please, I think failable `::new` methods are bad practice.

Gozala: I think it may make sense to do something like 

```rs
let invite = Invite::from(chain).claim(audience)?;
```

Where From impl would create Scoped version, but `.claim` would allow you to turn it into `Open` invite (I'd call it Redeemed personally). Then you could have `invite.delegate(audience)` if you want to turn it into delegation that can be stored.

rust/tonk-invite/src/lib.rs
```diff
@@ -0,0 +1,564 @@
...
+        DelegationChain::new(delegation)
+    }
+
+    async fn signer(seed: &[u8; 32]) -> Ed25519Signer {
+        Ed25519Signer::import(seed).await.unwrap()
+    }
+
+    #[cfg_attr(not(all(target_arch = "wasm32", target_os = "unknown")), test)]
+    #[cfg_attr(all(target_arch = "wasm32", target_os = "unknown"), wasm_bindgen_test)]
+    fn parse_rejects_non_url_input() {
```
Gozala: Can we please stick to the same pattern across all the crates, specifically:

1. Can we use `dialog_common::test` which does all the gating to run across targets so you don't have to.
2. We use `it_` style BDD naming convention.

Gozala (approved):
Suggested various improvements, but I leave it up to you whether you want to address or leave as is. Updating tests to use same convention would be only thing that I'd highly encourage and probably something LLM can easily carry out.

rust/tonk-invite/src/lib.rs
jackddouglas: `From` also wouldn't work since that needs to be infallible. we could do `TryFrom` but seems like the best option here is just to do `try_new`

PR #430, opened 2026-04-23
feat(shell): render repo space inside a same-origin sandboxed iframe

Gozala:
Adds a host/guest iframe bridge so each repository renders inside its own sandboxed `<iframe>` backed by the service worker instead of being printed directly into the shell's DOM. The JS shim now forwards every fetch to Rust, which decides per-request whether to route through the axum router or pass through to the network via `self.fetch`. Guest iframes get their URIs rewritten into `/api/repository/{repo}/...` before axum routes them, so iframe content can fetch relative to a virtual root without ever learning which repo it's in.

Summary of the moving parts:

- `TonkServiceWorker::on_fetch` now takes a `FetchEvent`, pulls `clientId` and `resultingClientId` (the latter via reflection — web-sys 0.3.85 has no binding for it), and decides via `route_for` whether to dispatch the request or pass it through. Pass-through is a direct `sw_fetch` with the SPA `/index.html` fallback for navigations — that behaviour lived in the JS shim before and now lives here alongside everything else.
- `router::host` serves the iframe's placeholder HTML at `/api/host/{host}/guest/{repo}/` and records a `clientId -> {host, repo}` binding on first load. Bindings hang off `TonkState.guests` so fetches from a known guest iframe can be rewritten without any extra handshake.
- `put_repository` now returns the existing `RepositoryInfo` on a same-shape `200 OK` when `If-None-Match: *` hits an existing repo, instead of `412 Precondition Failed`. reqwest's wasm client was aborting the 412 response body even though it carried valid JSON, which made first-load flakey; 200 is observationally equivalent for our caller and sidesteps the reqwest bug.
- `TonkShell::init` reads `X-Tonk-Client-Id` off the PUT response, publishes it through a `HostId` context signal, and `TonkSpace` uses it to build the iframe's `src`. The repository GET is gated on that signal so we don't race the service worker's activation with an unauthenticated fetch that would otherwise fall through to the dev server.

Isolation caveat worth naming: `allow-same-origin` is required for the sandboxed iframe to be SW-controlled at all, and it also gives the iframe full access to the parent's IndexedDB / DOM. The sandbox flags still prevent accidental top navigation, popups, forms, modals, and downloads, which is enough for content we author ourselves. Stronger isolation would require a second origin and a Lunet-style postMessage bridge, which is out of scope here.

### Description

_A few sentences describing the overall effects and goals of the pull request's commits.
What is the current behavior, and what is the updated/expected behavior with this PR?_

### Other changes

### Tested

### Related issues

### Backwards compatibility

### Documentation

---

## PR-Codex overview
This PR enhances the `tonk` application by implementing a host/guest iframe bridge, improving routing for guest clients, and refining the service worker's fetch handling to support SPA-style navigation and repository management.

### Detailed summary
- Added `space.css` for styling the new `.space` grid area.
- Updated `launcher.css` to modify grid layout.
- Implemented guest iframe handling in `host.rs`, including guest registration.
- Enhanced `api_router` to return shared application state.
- Improved service worker's fetch logic for SPA routing and client ID management.
- Updated `init` function to return Client ID for the UI context.
- Added error handling for repository creation and existing checks.

### DIFF
--- a/rust/tonk-ui/assets/service_worker.js
+++ b/rust/tonk-ui/assets/service_worker.js
@@ -24,29 +24,15 @@ self.onactivate = event => {
     log("Activated");
 };
 
+// Hand *every* fetch to the Rust side. The Rust worker decides
+// whether to route through its own handlers or pass the request
+// through to the network (which it does itself via `self.fetch`).
+// SPA-style 404 → index.html fallback and /api/* routing all
+// live in Rust now, so this shim is pure forwarding.
 self.onfetch = event => {
-    let request = event.request;
-    let url = new URL(request.url);
-
-    if (url.pathname.match(/^\/api/)) {
-        log(request.method, url.pathname);
-        event.respondWith(
-            (async () => {
-                return (await activateWorker()).onfetch(request);
-            })(),
-        );
-    // NOTE: Only intercept navigate as candidates for serving
-    // the index.html (in order to provide SPA-style routing)
-    } else if (request.mode === 'navigate') {
-        event.respondWith(
-            fetch(request).then(response => {
-                if (response.status === 404) {
-                    return fetch("/index.html");
-                }
-                return response;
-            }),
-        );
-    }
+    event.respondWith(
+        (async () => (await activateWorker()).onfetch(event))(),
+    );
 };
 
 // Background Sync API event handler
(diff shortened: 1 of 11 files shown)


Gozala:
This got incorporated into #444

PR #433, opened 2026-04-24
feat: webawesome

Gozala:
Integrates [webawesome](https://webawesome.com/) web components and switches UI to use them


Gozala:
merged as a part of #435

PR #440, opened 2026-04-28
feat(carry): push, pull, remote management, UCAN S3 sync

jackddouglas:
Wires carry up to dialog-repository's remote/upstream/sync surface

- `carry remote add <name> <url> [--subject DID]` registers a sync destination
- `carry remote {list,show,set-upstream,remove}`
- `carry push` / `carry pull` thin wrappers over `branch.push()` / `branch.pull()` against the configured upstream
- `join_cmd` auto-wires `origin` from an invite URL's `remote=` query parameter and runs an initial pull, so a freshly joined repo is immediately syncable
- `tests/sync_integration.rs` spins up `tonk-access-service::helpers` in-process and exercises the full invite -> join -> push <-> pull flow, including bidirectional sync

---

## PR-Codex overview
This PR introduces new features for managing remote repositories in the `carry` tool, including commands for adding, listing, and removing remotes, as well as implementing pull and push functionalities. It also adds integration tests for the UCAN-S3 sync path.

### Detailed summary
- Added `dialog-common`, `dialog-remote-s3`, and `dialog-remote-ucan-s3` dependencies.
- Introduced `pull_cmd` and `push_cmd` modules in `rust/carry/src/lib.rs`.
- Updated `REPO_NAME` to be `pub(crate)` in `rust/carry/src/site.rs`.
- Implemented `execute` functions for `carry pull` and `carry push` in `rust/carry/src/pull_cmd.rs` and `rust/carry/src/push_cmd.rs`.
- Created remote management commands in `rust/carry/src/bin/carry.rs`.
- Added remote commands for adding, listing, showing, setting upstream, and removing remotes in `rust/carry/src/remote_cmd.rs`.
- Implemented integration tests for UCAN-S3 sync in `rust/carry/tests/sync_integration.rs`.


rust/carry/src/bin/carry.rs
```diff
@@ -130,6 +130,89 @@ mod inner {
...
+            /// S3 access key ID (only for private s3:// remotes).
+            /// WARNING: persisted in plaintext inside .carry/.
+            #[arg(long = "access-key", value_name = "KEY")]
+            access_key: Option<String>,
+
+            /// S3 secret access key (only for private s3:// remotes).
+            /// WARNING: persisted in plaintext inside .carry/.
+            #[arg(long = "secret-key", value_name = "SECRET")]
+            secret_key: Option<String>,
+        },
```
Gozala, line 194: Feels like this should be an enum, but don't know if it's possible to express with args.

rust/carry/src/remote_cmd.rs
```diff
@@ -0,0 +1,328 @@
...
+            create.subject(did)
+        }
+        None => create,
+    };
+    create
+        .perform(&site.operator)
+        .await
+        .with_context(|| format!("failed to register remote '{}'", opts.name))?;
+
+    set_upstream(site, &opts.name).await?;
```
Gozala: I think we should follow git here and have a `--set-upstream` instead of implicitly setting one up on main branch.

rust/carry/src/remote_cmd.rs
```diff
@@ -0,0 +1,328 @@
...
+    eprintln!(
+        "Added remote '{}' and set it as the sync target.",
+        opts.name
+    );
+    Ok(())
+}
+
+/// Discover remote names by scanning the `.carry/{repo_did}/memory/` directory
+/// for `remote/*/address` entries. Returns sorted names.
+fn list_remote_names(site: &Site) -> Result<Vec<String>> {
```
Gozala, line 110: I think we should probably do what stable does now and store this information in the meta branch as facts. Alternatively we could consider extending dialog's capabilities to allow listing remotes instead. Either way this feels very brittle.

I also think that it's doing the wrong thing, remotes aren't actually stored under repo_did as far as I recall they should be stored under `.carry/memory/...` instead

rust/carry/src/remote_cmd.rs
```diff
@@ -0,0 +1,328 @@
...
+        }
+    }
+    names.sort();
+    Ok(names)
+}
+
+/// Format a [`SiteAddress`] as a human-readable URL string.
+fn format_site_address(addr: &SiteAddress) -> String {
+    match addr {
+        SiteAddress::S3(s3) => format!("s3://{}/{}", s3.endpoint(), s3.bucket()),
```
Gozala: Nit: I think this isn't actually incompatible with s3 UR scheme which is `s3://<bucket>/<key>`.

rust/carry/tests/sync_integration.rs
Gozala, line 1: Can you please use `dialog_common::test` integration tests that take care of provisioning servers and tearing them down so you don't have to do it manually in each test.

Here are some examples of it in action https://github.com/dialog-db/dialog-db/blob/fix/ucan-proof-chain-order/rust/dialog-repository/src/repository/branch/integration_tests.rs#L334

Gozala (approved):
Provided some feedback most of it minor, but I did spot potential bug in the way remotes are being listed which either assumes wrong paths or carry organizes files inside `.carry` different from what dialog does

rust/carry/src/bin/carry.rs
jackddouglas: it'd be nice for the pattern matching but i think it would end up looking something like this if we used subcommands:

```rs
carry remote add origin https https://access.example.com
carry remote add mirror s3 --endpoint https://s3.eu-west-1.amazonaws.com \
                           --region eu-west-1 --bucket my-bucket
carry remote add bobs origin https https://… --subject did:key:zBob
```

but we could make `RemoteAddOptions` an enum internally:
```rs
/// Options for registering a remote. The CLI layer normalises its flags
/// into one of these before calling [`execute`].
pub struct RemoteAddOptions {
    pub name: String,
    pub url: String,
    /// Subject DID at the remote. `None` means "use my own repo DID",
    /// which is the common case (syncing your own repo to your own
    /// bucket). `Some(did)` is for cross-repo pulls (e.g. Bob pulling
    /// Alice's data).
    pub subject: Option<String>,
    pub s3_endpoint: Option<String>,
    pub s3_region: Option<String>,
    pub s3_bucket: Option<String>,
    pub s3_access_key: Option<String>,
    pub s3_secret_key: Option<String>,
}
```

PR #445, opened 2026-05-06
feat: tonk concept component

Gozala:
### Description

Adds `<tonk-concept>` — a custom element that subscribes to a tonk-worker query and renders each match into an author-supplied template — plus a SPA route at `/space/:space/branch/:branch/concept/:source` that demos it.

The reactor work landed earlier gave us a `/query` SSE endpoint, but nothing on the page consumed it. `<tonk-concept>` makes the live-update loop visible: drop the tag into any HTML, write the row markup with `{field}` interpolation, and rows update in place as the branch changes. No framework wrapper, no signals, just markup

The element discovers the concept's shape itself (one Phase-1 lookup) before opening the live subscription, so `source="person"` is the only thing the author has to specify. That required exposing `concept:` as a queryable concept, which is the other half of this PR.

### Example

A concept defined in tonk-notation, asserted on the `home` branch:

```yaml
attribute! person-name:
  the:         xyz.tonk.person/name
  as:          Text
  cardinality: one

attribute! person-age:
  the:         xyz.tonk.person/age
  as:          UnsignedInteger
  cardinality: one

concept! person:
  with:
    name: .person-name
    age:  .person-age
```

Rendered live with one element. The `<ul>` is static chrome; the `<template>` body is cloned per match into one `<li>` row:

```html
<tonk-concept source="person">
  <ul>
    <template>
      <li><b>{name}</b> is {age}</li>
    </template>
  </ul>
</tonk-concept>
```

Each asserted person becomes one `<li>`; updates re-render in place; removed entities drop. `<tonk-concept source="person?name=Alice">` filters the live subscription to one row.

### Tested

- Native unit tests for the pure pieces (segment + source parsers, query builders).
- Wasm browser tests for the templating pipeline (insert/update/remove/append, both template-detection paths, attribute substitution, write-deduping).
- Wasm router test for `reactor.shutdown()` ending SSE subscribers on SW upgrade.
- Manual exercise via Chrome MCP at `/space/home/branch/main/concept/person` and `?name=Alice`.


Gozala:
This got folded into staging branch

PR #481, opened 2026-06-02
feat: optional concept attributes via maybe: block

Gozala:
## Why

Concepts could only declare required attributes: every field of a concept had to be present on every assertion, and a `concept:` query excluded any entity missing any field. That makes optional data (a nickname, a bio) impossible to model without splitting concepts. This adopts dialog-db's optional-attribute support ([dialog-db#346](https://github.com/dialog-db/dialog-db/pull/346)) and threads it through notation, the analyzer, the evaluator, storage, and the LSP.

## Approach

A concept body gains a second field block: `with:` for required fields, `maybe:` for optional ones. Both are symmetric (bare references and inline definitions alike); `maybe:` simply implies optional. They merge into dialog's single `with` map, each field carrying a per-field `optional` flag.

```
concept!: &person
  with:
    name: person-name        # required
  maybe:
    nickname:                # optional
      as: text
```

- **Stop requiring optionals on assertions:** the completeness check counts only required fields, so omitting an optional on a fresh entity is no longer an error. A field declared in both blocks is rejected (`E_DUPLICATE_CONCEPT_FIELD`), surfaced through the LSP.
- **Set-widening on query:** dialog's planner derives optional resolution from the descriptor's per-field flag, so a `concept:` query returns entities that lack an optional field with that field simply omitted, instead of filtering them out. `Match::lookup` now returns `Binding` (Present/Absent); Absent bindings are dropped from frames, conclusions, and head facts.
- **Storage:** optionality persists as a boolean marker `dialog.concept.optional/{field}` alongside the existing `dialog.concept.with/{field}` link. Required-field storage is byte-identical, so existing branches and concept identities are unaffected.

## Heads-up for review

The `dialog-*` deps point at the **`feat/type-inference-v2` branch**, not a tag, because the upstream PR hasn't merged. This needs repinning to a tag/commit once it does — flagged in `plan/optional-attributes.md`. That branch also brings rule-level type inference, which legitimately rejected one pre-existing test fixture with a latent `?this` type conflict (fixed here).

`plan/optional-attributes.md` documents the full design, the deviations found against the real branch source, and the test coverage (plus three low-risk known gaps).

---

## PR-Codex overview
This PR primarily focuses on enhancing the handling of required and optional fields in the `concept!` declarations, updating type handling, and improving error messages related to field types. It also revises the use of the `tonk:_` wildcard model in various contexts.

### Detailed summary
- Changed `as: text` to `as: unsigned-integer` in `demo.yaml`.
- Updated error messages to reference `tonk:_` instead of `_:_`.
- Improved handling of required and optional fields in `concept!`.
- Introduced checks for duplicate field declarations across `with:` and `maybe:`.
- Enhanced type coercion and error handling for scalar values.
- Adjusted tests to ensure proper behavior with optional fields.

> The following files were skipped due to too many changes: `rust/tonk-evaluator/src/evaluate.rs`, `rust/tonk-schema/src/concept.rs`, `rust/tonk-analyzer/src/analyzer.rs`, `plan/optional-attributes.md`


Gozala:
## Follow-up: optional query set-widening is blocked on a dialog-db planner bug

While verifying the headline behavior (a `concept:` query returning entities that lack an optional field), I found a bug in dialog-db's concept-query planner, not in this integration.

**Symptom:** a `this`-unbound query like `person:` over a concept with required `name` + optional `age` returns only the people who *have* an `age`. The set-widening (`Absent` fallback) silently drops the rest.

**Root cause:** the planner leads the unbound scan with the alphabetically-first field. When that lead is optional, its `Absent` fallback is suppressed while `this` is unbound, so missing-optional entities never produce a row. `age` < `name`, so it triggers; an optional field sorting *after* the required field works fine — which is why the basic case passed initially.

Minimal dialog-only repro + fix tracked upstream: **dialog-db#348** (https://github.com/dialog-db/dialog-db/pull/348).

**What this commit adds here:**
- A real tonk-side fix: reconstruction queried concept claims with `is` typed as `Entity`, which dropped the Boolean `dialog.concept.optional/*` markers, so the optional flag never round-tripped from the branch. Now uses a separate Boolean-typed marker query.
- `it_set_widens_optional_field_in_query` — passes (optional sorts after required).
- `it_set_widens_body_derived_entities_with_bare_query` — `#[ignore]`d, the real-world `person:` repro, blocked on dialog-db#348.
- `dialog_repro_optional_field_sorted_first_drops_rows` — `#[ignore]`d, minimal dialog-level repro from this side.
- `it_reconstructs_optional_flag_from_branch` (in tonk-schema) — round-trips `is_optional` through persistence + reconstruction.

The two ignored tests should be un-ignored once dialog-db#348 lands and we repin off the branch.

Gozala:
## Regrafted onto latest staging + moved to dialog `feat/narrowing-diagnostics`

Rebased onto current staging and swapped the dialog deps from `tonk-2026-06-22` to the `feat/narrowing-diagnostics` branch (the operator-IR + type-inference query-engine rework, dialog PRs #347/#350/#351/#352).

**The optional set-widening bug is fixed by that rework.** A `this`-unbound query (`person:`) over a concept with an optional field that sorts alphabetically first no longer drops entities lacking it. Both regression guards are now un-ignored and pass:
- `it_set_widens_optional_field_sorted_before_required` (dialog-level)
- `it_set_widens_body_derived_entities_with_bare_query` (the original `person:` repro)

The dialog-side repro PR (dialog-db#348) can be closed/superseded by the rework.

**API adaptations for the new branch:** `ConceptDescriptor::operands()` → `required_operands()` (head-fact emission appends optional keys back, Absent skipped); `dialog-reactor`'s `Decode` reads the descriptor via `Descriptor<ConceptDescriptor>` (derive dropped `From<Query>`); two test fixtures updated for the now-rejected empty concept.

**One unrelated regression surfaced** by the dialog upgrade's strict Entity/Text typing: the `_:_` default-model view sentinel is stored as Text but queried as Entity, and `_:_` isn't a valid entity URI. Tracked as BUG-16; its slide test is `#[ignore]`d pending a sentinel-representation decision. Not an optionals issue.

Full workspace is green (fmt + clippy clean).

PR #508, opened 2026-06-11
refactor(inspector): make tonk-inspector a host-driven notebook

Gozala:
The inspector grew its own opinions about the worker: it fetched the repository directly, owned a branch selector, header, push/pull sync, and remotes views. That duplicated machinery the rest of the app already gets for free from the outer `<tonk-host>`, and it meant the inspector couldn't be dropped into a board tile without dragging all that chrome along.

This reshapes `<tonk-inspector>` into a plain host consumer, the way `<tonk-display>` works: a simple notebook of `<tonk-code>` cells that evaluate and render results, and nothing else.

## Approach

- The notebook dispatches `tonk-evaluate` on its editor element and lets the surrounding routing context (`<tonk-repository>` / `<tonk-branch>` under `<tonk-host>`) annotate space + branch and perform the IO. No `fetch`, no `api::*`, no branch state of its own.
- The host's evaluate path gains a `transact` flag so the inspector can keep its two modes: a dry-run preview on every clean LSP diagnostics frame, and a committing submit on Shift+Enter / the play button.
- Branch-aware LSP keeps working. The language server derives `(repo, branch)` from the buffer URI, so the notebook computes `tonk-buffer:///<repo>/<branch>/scratch-{id}` from the same routing ancestors the host annotator reads — the inspector still holds no branch state; it only reads ambient DOM context.
- The Repository-side sync apparatus in the shared `BranchRow` (sync state, pull/push, sync-status polling, upstream badges, branch-channel watches) became dead once the inspector stopped using it — the only remaining caller is the profile route, which never had an upstream — so it's removed. `profile.rs` keeps working through the Profile path.

Dropped on purpose: the eval-time analyzer-diagnostic push (the host's evaluate error carries no positional range). The editor's own on-change LSP squiggles are unaffected; eval errors surface in the failure callout instead.

`core.yaml`'s inspector concept keeps its `source` field as the entity anchor; the element simply ignores it, so no stdlib reseed is needed.

Verified: wasm build clean (zero warnings), native test-compile clean, `cargo fmt` + clippy clean on the changed crates. Not yet exercised in a live browser.

---

## PR-Codex overview
This PR primarily focuses on refactoring the `tonk` application by modifying the handling of profile branches, enhancing the `evaluate_url` function, and improving the `TonkInspector` component.

### Detailed summary
- Removed `BranchOwner` from `TonkProfile`.
- Updated the `evaluate_url` function to include a `transact` parameter.
- Enhanced documentation for clarity on the `evaluate` function.
- Improved the `TonkInspector` with new structure and functionality.
- Removed unused sync state management code.
- Added profile-specific editor source handling in `BranchRow` and `BranchCell`.


Gozala:
Merged into #516

PR #524, opened 2026-06-19
fix(slide): trust OS CA store for TLS so slide works behind MITM proxies

goblinoats:
> 🤖 **Automated fix** authored by Claude Code from issue **BUG-12 — "Cannot use slide in Claude sandbox environments"** in the `TonkSpotLaunchList` space.

## Problem

`slide` remote operations (`pull` / `push` / `status` / `share`) fail with a transport error in any TLS-intercepting environment — corporate proxies, Zscaler, and notably the **Claude Code sandbox egress gateway**:

```
error: Failed to fetch from remote: ... Transport error: error sending request
for url (https://staging.tonk.xyz/ucan/)
```

`join` / `init` work because they're local; anything touching the remote dies during the TLS handshake.

## Root cause

The remote transport runs through dialog-db's `dialog-remote-ucan-s3` and `dialog-remote-s3` crates, which build their HTTP client with `reqwest::Client::new()`:

- `dialog-remote-ucan-s3/src/site.rs:33`
- `dialog-remote-s3/src/s3/permit.rs:58`

dialog-db's workspace configures reqwest with `features = ["rustls-tls", "json"]`. reqwest's `rustls-tls` = **embedded webpki/Mozilla roots only** — it never consults the OS trust store or `SSL_CERT_FILE` / `SSL_CERT_DIR`. So when the egress gateway presents a cert signed by a private CA that *is* installed system-wide (and trusted by curl/git/python), slide rejects it. It's a client-side cert-trust rejection, not a network block.

## Fix

Add a reqwest `0.12` dependency to the `slide` crate — pinned to the same version dialog-db resolves — enabling the `rustls-tls-native-roots` feature:

```toml
reqwest = { version = "0.12", default-features = false, features = ["rustls-tls-native-roots"] }
```

Cargo unifies features across the dependency graph, so dialog-db's reqwest client picks up native roots too. TLS root features are **additive**: reqwest loads webpki **and** native roots into a single store (verified in `reqwest-0.12.28/src/async_impl/client.rs` — two independent `#[cfg]` blocks, both default-on). So this is a strict **superset**:

- Every certificate that validated before still validates (webpki roots remain active).
- CAs in the OS trust store — including the sandbox gateway CA — now validate too.

**Non-breaking for normal users**: on an ordinary machine the OS store contains the standard public CAs, exactly like curl/git. The change only ever *grows* the trusted set.

## Verification

- `cargo tree -p slide -i reqwest -e features` → single `reqwest v0.12.28` with both `rustls-tls-native-roots-no-provider` and `rustls-tls-webpki-roots-no-provider` active.
- `cargo build -p slide` → clean.
- `rustls-native-certs v0.8.3` now present in the slide dependency tree alongside `webpki-roots v1.0.6`.

## Follow-up (durable fix)

This is a **local stopgap**. The permanent fix belongs upstream in dialog-db: add `rustls-tls-native-roots` to its workspace reqwest features, cut a new tag, and bump the `tag = "tonk-2026-06-16"` pins in the root `Cargo.toml`. Once that ships, this line can be dropped.

---

## PR-Codex overview
This PR focuses on updating dependencies in the `Cargo.toml` and `Cargo.lock` files, particularly around the `reqwest` library, to improve compatibility with TLS configurations and remove unnecessary dependencies.

### Detailed summary
- Added `reqwest` as a workspace dependency for compatibility with TLS.
- Updated `reqwest` version to `0.12` with specific features enabled.
- Removed several unused packages from `Cargo.lock`.
- Adjusted dependencies to unify features across packages.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -237,28 +237,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
-name = "aws-lc-rs"
-version = "1.15.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7b7b6141e96a8c160799cc2d5adecd5cbbe5054cb8c7c4af53da0f83bb7ad256"
-dependencies = [
- "aws-lc-sys",
- "zeroize",
-]
-
-[[package]]
-name = "aws-lc-sys"
-version = "0.37.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5c34dda4df7017c8db52132f0f8a2e0f8161649d15723ed63fc00c82d0f2081a"
-dependencies = [
- "cc",
- "cmake",
- "dunce",
- "fs_extra",
-]
-
-[[package]]
 name = "axum"
 version = "0.8.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -529,16 +507,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "47b26a0954ae34af09b50f0de26458fa95369a0d478d8236d3f93082b219bd29"
 dependencies = [
  "find-msvc-tools",
- "jobserver",
- "libc",
  "shlex",
 ]
-
-[[package]]
-name = "cesu8"
-version = "1.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6d43a04d8753f35258c91f8ec639f792891f748a1edbd759cf1dcea3382ad83c"
 
 [[package]]
 name = "cfg-if"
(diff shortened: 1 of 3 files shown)


goblinoats:
Fixed the failing **Lints** check (`sharedWorkspaceDeps`). It enforces that every crate dependency is declared via `[workspace.dependencies]`; the first push pinned `reqwest` inline in `rust/slide/Cargo.toml`, which violated that.

Reworked so the spec lives in the workspace: the workspace `reqwest` now pins `0.12` (matching dialog-db's transport crates) with `rustls-tls` + `rustls-tls-native-roots` + `json`, and slide references it via `reqwest = { workspace = true }`. Same fix, lint-compliant.

Notes:
- reqwest gates its TLS deps behind `cfg(not(target_arch = "wasm32"))`, so the added features are a no-op for `tonk-ui`'s wasm build. `tonk-ui` only moves `0.13` → `0.12` (the version already in the tree via dialog-db).
- Dropping the second reqwest version removed its exclusive `aws-lc-sys`/`cmake` crypto tree from the lockfile (~220 fewer lines).
- Verified locally: `slide` builds; `tonk-ui` builds for native **and** `wasm32-unknown-unknown`; `cargo-enforce-shared-workspace-deps` (the exact tool CI runs) passes.

PR #561, opened 2026-07-05
feat: content-addressed blob support

jackddouglas:
Adds blob support (M1 local + M2 sync): images and other binary data can now be
stored in a space, referenced from claims, and replicated through push/pull.

## What you can do

    tonk blob add photo.png        # ingest, print blob:<hash>; asserts content-type + name facts
    tonk blob ls                   # blob index: hash, size, content-type
    tonk blob cat blob:<hash>      # bytes to stdout, lazily fetched from the remote on miss

    photo!:
      this: id:vacation
      image: blob:<hash>           # blobs are ordinary entities in notation

`tonk push` ships newly referenced blob bytes to the remote before publishing the
revision; peers that pull get the blob *index* (hash + size) immediately and
hydrate bytes on first read, digest-verified.

## Design

A blob is identified by the BLAKE3 hash of its content and referenced as a
`blob:<base58-hash>` entity, so ordinary facts carry extrinsic metadata
(content-type, name — see the new `tonk:blob` concept in the standard library).
Intrinsic metadata (size) lives in the blob index, a fifth key ordering in the
same prolly tree, which makes "which blobs does the remote lack" a plain tree
differential — no side ledger. Bytes never enter the tree. Full write-up:
`docs/superpowers/specs/2026-07-05-blob-support-design.md` and dialog-db's
`notes/blob-replication.md`.

## Upstream

Depends on dialog-db tag `tonk-2026-07-06` (feat/blob-sync): repository-level
`write_blob` / `read_blob` / `blob_size` / `list_blobs`, blob-shipping push,
blob effect routing (native fs, in-memory for volatile spaces, OPFS via a
dedicated `Space` blob field on web), and UCAN authorization for blob
read/import. The access service itself needed no changes — blob capabilities
flow through the existing `/ucan/` route.

---

## PR-Codex overview
This PR introduces a new feature for handling content-addressed blobs in the `tonk` application, allowing users to add, retrieve, and list blobs via HTTP. It also enables rendering blobs as inline images in the web frontend.

### Detailed summary
- Added `blob` module in `rust/tonk-cli/src/lib.rs`.
- Implemented HTTP route to serve blob bytes in `rust/tonk-worker/src/router/blob.rs`.
- Created `blob` commands in the CLI for adding, retrieving, and listing blobs.
- Enhanced `tonk-display` to render blobs as `` elements.
- Updated tests for blob functionalities in `rust/tonk-cli/tests/blob.rs` and `rust/tonk-worker/src/router/blob.rs`.
- Documented blob usage in `guide/src/reference.md`.


jackddouglas:
## Web inline rendering for image blobs

Adds the missing web-display path so a content-addressed image blob renders inline in the frontend (the substrate — storage, sync, the \`tonk:blob\` concept, the \`tonk blob\` CLI — already landed on this branch). Scope: **images only, inline-in-views only.**

**Commits**
- `feat(tonk-worker)`: new route `GET /api/repository/{repo}/branch/{branch}/blob/{entity}` streaming a blob's bytes from the branch blob store, `Content-Type` from the `xyz.tonk.blob/content-type` fact (mirrors the sibling `host::guest` handler). Rejects non-blob entities up front via `Entity::blob_hash()`.
- `feat(tonk-display)`: when a `<tonk-display>`'s `model` is `tonk:blob`, it mounts a single `` whose `src` is that route, built from the host's `with` (`{branch}@{repo}`) + `entity`. Keyed on the model string (not a projected `type` field) deliberately — adding `type` to the base view predicate would force a required fact onto every existing view.
- `docs(guide)`: note the inline-render usage.
- `fix(tonk-worker)`: log content-type query errors instead of silently defaulting; add a test for the non-blob-entity rejection guard.

**Tests** — worker route verified in a wasm service-worker context (incl. the rejection guard); the `tonk:blob`→`` dispatch verified live in Chrome. Native clippy clean on both crates.

**Not in this change (follow-ups):** PDFs/other MIME types, a dedicated blob route/page, tonk-render (headless) parity, blob streaming (bytes are buffered), and an in-browser blob-upload route. One manual check still open: rendering a real image blob end-to-end in the running frontend (needs a synced space); the component-level tests cover both halves of the seam.

_Implementation plan: `docs/superpowers/plans/2026-07-08-web-blob-images.md`._

PR #576, opened 2026-07-09
fix(worker): read lock on /evaluate, serialize its commit per branch

Gozala:
Follows up the June 22 lock rework. That pass dropped `/transact`, `/sync`, `/pull` and `/push` to a read lock on `TonkState` and moved commit serialization onto a per-branch lock — but `/evaluate` was left holding the write lock across parse, evaluate, and commit.

Because `tokio`'s `RwLock` is write-preferring, one write-lock holder stalls every new reader. The language server holds a `TonkState` read guard for the length of a `didChange` (and holds the global `Mutex<Server>` while it does), and the editor auto-fires a dry-run evaluate on every fresh diagnostics frame. So typing contended with itself: an auto-evaluate took the write lock, and the next LSP request queued behind it while holding the server mutex, head-of-line blocking every other cell and tab.

The catch is that `/evaluate` can't just drop to a read lock. Its commit is a *dialog* `Transaction::commit()` (it needs the raw transaction to thread through the evaluator), which CASes against the head snapshot and never retries — unlike `/transact`, which commits through the reactor builder that takes the per-branch transactor lock and refresh-retries on a moved head. The global write lock was the only thing serializing evaluate-vs-evaluate and evaluate-vs-transact commits; removing it alone would trade contention for version-mismatch failures.

So the committing path now takes `session.transactor()` — the same per-branch lock `/transact` uses — held from before the head snapshot through the publish. A dry run (`transact=false`), which is the hot auto-evaluate path, takes no lock at all.

Same read-lock downgrade for the claim routes and the read-only CSV export, plus the `sync` status stamps (removing a pointless drop-read-then-take-write upgrade). `pull`, `push` and `import` deliberately keep the write lock: they commit through the unlocked dialog path and, unlike `sync`, have no refresh-and-retry loop, so excluding writers is what keeps a racing commit from failing them — and none is on a hot path.

Verified: `cargo fmt` + `clippy -D warnings` clean; native suite green; the tonk-worker wasm suite green in headless Chrome, including the two `/evaluate` tests that commit through the new transactor lock. (Two profile-identity tests fail under the full wasm run on `staging` today, independent of this change — they pass in isolation and fail identically on the base.)

---

## PR-Codex overview
This PR focuses on improving the locking mechanism in the `tonk-worker` and `dialog-reactor` to enhance concurrency and reduce contention during read and write operations.

### Detailed summary
- Changed `write()` locks to `read()` locks in multiple locations to optimize performance.
- Made `COMMIT_RETRY_LIMIT` and `is_head_moved` public for broader accessibility.
- Updated documentation comments to clarify locking behaviors and transaction handling.
- Refactored `evaluate_on_branch` to handle retries for head movements during commits.

### DIFF
--- a/rust/dialog-reactor/src/lib.rs
+++ b/rust/dialog-reactor/src/lib.rs
@@ -60,7 +60,7 @@ pub use subscription::{QueryHash, Subscriber, SubscriptionPoll, SubscriptionRefe
 /// elements) can deserialize without depending on this crate.
 pub use tonk_schema::conclusion::Conclusion;
 pub use tonk_schema::query::Query;
-pub use transaction::{Commit, TransactionBuilder};
+pub use transaction::{COMMIT_RETRY_LIMIT, Commit, TransactionBuilder, is_head_moved};
 
 /// A reactive layer over dialog branches. Owned by the consumer's
 /// application state (e.g. the worker's `TonkState`).
--- a/rust/dialog-reactor/src/transaction.rs
+++ b/rust/dialog-reactor/src/transaction.rs
@@ -242,13 +242,20 @@ impl Commit<'_> {
 /// How many times a reactor commit refreshes-and-retries when a sync advanced
 /// the head under it. Transactions serialize via the transactor lock, so the
 /// only racer is sync; one retry settles it in practice.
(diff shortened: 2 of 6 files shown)


Gozala:

PR #589, opened 2026-07-12
fix(reactor): serve a subscriber that attaches during a poll

Gozala:
## Why

A `<tonk-display>` sometimes gets stuck on `data-state="loading"` (the never-ending spinner) — most visibly a nested `<tonk-site model="tonk:site">` whose `tonk:site` stamp races the guest's subscription. The subscription is established but never receives its first frame.

## Cause

`SubscriptionPoll::perform` decided whether to build a `Frame::Snapshot` from an `any_pending` flag captured **before** the engine poll's `await`. A subscriber that attaches while that poll is running (a very common race: the guest subscribes at almost the same moment the `tonk:load` claim stamps `tonk:site` and triggers the poll) is not counted in that flag, so no snapshot is built for it. At fan-out it's `Pending` but `snapshot_bytes` is `None`, so it's silently kept unserved. The comment claimed "the next scheduled poll re-serves it" — but on a quiescent branch (nothing else writes), no next poll is ever scheduled, and the subscriber hangs forever.

## Fix

Decide delivery per subscriber **at fan-out**, under the map lock, instead of off a pre-poll flag. The snapshot is projected from the just-polled engine results and serialized lazily — built only if a subscriber is actually `Pending` at delivery time. A subscriber that attached during the poll is therefore served its first snapshot in the same poll that ran because of it.

The fan-out is extracted into `Subscription::deliver` / `fan_out`, and four regression tests cover it directly (Pending-beside-Established, multiple Pending, delta-to-Established + snapshot-to-Pending in one round, closed-channel pruning) — the delivery logic that had the race is now exercised without needing a live dialog engine.

---

## PR-Codex overview
This PR enhances the subscription delivery mechanism in the `dialog-reactor` by ensuring that subscribers receive the correct data based on their status, particularly addressing issues with pending subscribers not receiving timely snapshots.

### Detailed summary
- Removed the `Status` import from `super::state`.
- Simplified logic to determine if subscribers need a snapshot during polling.
- Introduced `Subscription::deliver` method to handle delivery logic.
- Added a `fan_out` function to manage subscriber notifications.
- Implemented lazy serialization for snapshot bytes.
- Improved handling of subscriber states during delivery.
- Added tests to ensure correct behavior for pending and established subscribers.

### DIFF
--- a/rust/dialog-reactor/src/subscription/reference.rs
+++ b/rust/dialog-reactor/src/subscription/reference.rs
@@ -17,7 +17,7 @@ use crate::BranchState;
 use crate::env::SelectProvider;
 use crate::{Conclusion, Frame};
 
-use super::state::{QueryHash, Status};
+use super::state::QueryHash;
 
 /// Names a subscription within a branch. Built from
 /// [`BranchSession::subscription`].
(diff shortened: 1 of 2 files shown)


Gozala:

PR #591, opened 2026-07-12
perf(sync): defer the sync drain while a page is loading

Gozala:
## Why

The background sync drain runs its fetch + prolly-tree merge on the single service-worker thread. The reactor already keeps sync lock-light (the network fetch is lock-free, the apply holds the transactor lock only for microsecond cell-writes, and a no-op pull does no subscription work) — so this is not lock contention. What remains is **thread occupancy**: when a sync finds an actual update, the merge CPU work runs on the one SW thread and competes with an actively-loading page's data-plane request processing. This is worst for a page booting against an already-warm worker (a new tab, or a reload), where the cold-start timing offers no protection.

## How

The sync scheduler now counts in-flight `/api/*` requests and holds the drain off while any are outstanding. A page actively booting keeps its data-plane requests in flight, so the drain waits until the load settles, then runs. Details:

- Only `/api/*` requests count — static-asset fetches never touch the reactor.
- An RAII guard decrements the count on any request outcome (success, error, or panic).
- Time-bounded by `SYNC_LOAD_DEFER_MS` (2s), comfortably longer than a real load (measured sub-500ms) but short enough that a long-lived SSE stream or a genuinely hung request can't defer sync indefinitely.

Five scheduler unit tests cover the new behavior (defer while loading, drain once settled, stop deferring past the cap, concurrent-load counting).

Note: measurement on staging showed the load itself is already fast (sub-12ms per boot request on a warm SW), so this is a targeted guard against the update-during-load case rather than a fix for a large observed regression.

---

## PR-Codex overview
This PR introduces enhancements to the `SyncScheduler` in `worker.rs`, adding functionality to manage data-plane requests more effectively. It implements tests to ensure that the sync drain correctly handles loading states and request counts, preventing delays caused by in-flight requests.

### Detailed summary
- Added four tests to verify behavior of `SyncScheduler` during loading states.
- Implemented `enter_loading` method to track active data-plane requests.
- Introduced `is_loading` method to check if a request is in flight.
- Enhanced `should_drain` method to consider loading states.
- Added `LoadingGuard` struct to manage the count of in-flight requests.
- Defined `SYNC_LOAD_DEFER_MS` constant to limit how long requests can defer sync.

### DIFF
--- a/rust/tonk-worker/src/worker.rs
+++ b/rust/tonk-worker/src/worker.rs
@@ -506,6 +506,68 @@ mod route_for_tests {
         );
         s.end_drain();
     }
+
+    #[dialog_common::test]
+    fn it_defers_a_drain_while_a_page_is_loading() {
+        let s = SyncScheduler::default();
+        let t1 = s.next(0.0);
+        // A data-plane request is in flight (a page booting): even the latest
+        // ticket at its debounce edge must NOT drain — sync yields the thread.
+        let _guard = s.enter_loading(0.0);
+        assert!(
+            !s.should_drain(t1, SYNC_DEBOUNCE_MS as f64),
+            "an in-flight data-plane request defers the drain",
+        );
+    }
+
+    #[dialog_common::test]
+    fn it_drains_once_the_load_settles() {
+        let s = SyncScheduler::default();
+        let t1 = s.next(0.0);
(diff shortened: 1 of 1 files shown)


Gozala:

PR #598, opened 2026-07-13
fix(ci): unbreak crate downloads; upgrade worker + wasm-bindgen

Gozala:
Every branch started failing `Lints` at once, with no code change to blame — including branches that touch no dependency files at all.

```
cargo-enforce-shared-workspace-deps-…-vendor-staging.drv failed
Failed to fetch file from https://crates.io/api/v1/crates/equivalent/1.0.2/download
Status code: 403
```

A different crate each run, so it read like flakiness. It isn't.

## Cause

**crates.io now 403s requests that send no `User-Agent`.** Verified by hand:

```
curl -sI https://crates.io/api/v1/crates/anyhow/1.0.100/download           -> 403
curl -sI -A "anything" https://crates.io/api/v1/crates/anyhow/1.0.100/download -> 200
```

Our `nixpkgs` was pinned to **2026-02-08**, whose `fetch-cargo-vendor-util.py` sets no UA. Upstream has since fixed it:

```python
session.headers["User-Agent"] = "nixpkgs-fetchCargoVendor/2 (https://github.com/NixOS/nixpkgs)"
```

That is the whole failure. It broke everything simultaneously because it is a change on crates.io's side, not ours.

Worth noting: the download that fails is not one of *our* dependencies. It is nixpkgs vendoring the crates for `cargo-enforce-shared-workspace-deps` — a lint tool. Nothing in any of the failing PRs could cause or fix it.

## Fix

Bump `nixpkgs` (2026-02-08 → 2026-07-11). Only that input moves — the lockfile diff is 3 lines.

## Verification

I could not trust local `nix flake check` here (my store hit a disk-full and has invalid paths), so CI on this PR is the real test. If `Lints` goes green, the diagnosis holds and every other branch unblocks with it.

Timeline that isolates it — `stable` was green at 19:40 UTC, and every run after that fails identically, including `feat/agent-build`, which is unrelated to any of my work.

---

## PR-Codex overview
This PR focuses on updating dependencies, specifically the `wasm-bindgen` ecosystem, and making changes to how `Response` objects are handled in the Rust codebase. It also includes minor version updates for various packages.

### Detailed summary
- Updated `wasm-bindgen` and related packages from version `0.2.108` to `0.2.126`.
- Changed handling of `web_sys::Response` to use `JsValue::from(resp)` instead of `resp.into()`.
- Updated `wasm-streams` from version `0.4.2` to `0.6.0`.
- Updated multiple dependencies, including `rand`, `worker`, and `http-body`.
- Updated `dialog-db` dependencies to tag `tonk-2026-07-14-3`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -479,9 +479,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.2.66"
+version = "1.2.67"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f5d6cac793997bd970000024b2934968efe83b382de4fdcf4fcb46b6ee4ad996"
+checksum = "e17dd265a7d0f31ef544e1b20e03add05d3b45b491b633b10d67145d2acc1a38"
 dependencies = [
  "find-msvc-tools",
  "jobserver",
@@ -907,7 +907,7 @@ dependencies = [
 [[package]]
 name = "dialog-artifacts"
 version = "0.1.0"
-source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-13-2#ec3abffbbc093cae07a83c34c3c240e3bedb6863"
+source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-14-3#4cbdee5a4e7289f665cfc5103a521acbc73f1204"
 dependencies = [
  "arrayref",
  "async-stream",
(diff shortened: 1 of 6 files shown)


Gozala:
Landed via #586, which merged this branch to unblock its own CI. Verified in `staging`: `worker` 0.8.5, `wasm-bindgen` 0.2.126, `wasm-streams` 0.6, dialog pinned at `tonk-2026-07-14-3`, and nixpkgs on 2026-07-11.

Also in `stable` via #597.

PR #613, opened 2026-07-16
feat: declarative blob renderer (seeded tonk:blob media view + file card)

jackddouglas:
Blob rendering becomes a seeded, user-overridable view instead of a hardcoded image path. Follows #561 (image blob rendering) and #590 (blob upload component).

Spec: `docs/superpowers/specs/2026-07-15-declarative-blob-renderer-design.md`
Plan: `docs/superpowers/plans/2026-07-16-declarative-blob-renderer.md`

## What changed

- **core.yaml seeds `view!: tonk:blob/media-view`** (`model: tonk:blob`) carrying a `tonk-blob-media` web component inline via `<tonk-component>`'s inert script holder. `image/*` content types fetch bytes through the relayed `window.fetch` and mount an object-URL `` (native `src=/api/…` loads bypass the relay inside the sealed guest); every other type renders a file card (name, content-type badge, lazy download). Re-asserting the view with the same `this:` replaces the whole renderer — that is the override mechanism, and it is proven live.
- **tonk-display demotes the native blob `` to a default slide**: the `model == "tonk:blob"` early return moves into the empty-view-frame path and registers as slide `"__blob__"`, so any resolved view (the seeded one, or a user override) replaces it through ordinary slide reconciliation. Old/unseeded branches (including slide branches) keep working via the fallback. The fallback survives empty→view→empty round trips (stale view slides are swept, the img lookup is scoped to direct children).
- **The `xyz.tonk.blob/name` fact always lands on upload**: the worker `POST …/blob` route defaults it (header wins, an existing fact is preserved on headerless re-upload, entity string as last resort) and `tonk blob add` likewise. The concept query behind the seeded view matches only rows with every field present, so a nameless blob would never render. `BlobUploadResponse.name` is now always present.
- **The sealed guest sandbox gains `allow-downloads`** (tonk-portal). Live verification proved the download click silently produced nothing inside the opaque guest while the identical flow worked top-frame. The flag applies at every portal nesting level deliberately: sandbox flags intersect across nesting (the live topology is a doubly-nested guest), and the guest already executes arbitrary synced branch views, so a nested portal is not a lower-trust context. `allow-same-origin` and `allow-top-navigation` remain withheld.
- The component JS is defensive: object-URL lifecycle is generation-guarded against stale fetches and mid-fetch disconnects, failed fetches clear the render key so a later frame retries, no-op re-renders don't starve in-flight fetches, and user data reaches the DOM via `textContent`/property assignment only.

## Live verification (real dev stack, headless Chrome)

- PNG uploaded through the real `<tonk-upload>` renders through the seeded view: `` inside `tonk-blob-media`'s shadow root with a `blob:` URL — not the native fallback.
- PDF renders the file card; Download saves a byte-identical file (after the `allow-downloads` fix; before it, the click was silently blocked in the guest).
- Re-asserting `tonk:blob/media-view` with a modified template overrides the renderer on reload.

## Tests

`test:native:debug` 918 passed, `test:web:debug` 916 passed (includes a new round-trip test: empty frame → fallback img; view resolves → replaces it; view retracted → fallback remounts without capturing the override view's img), `standard_library` lowering 10/10, fmt + workspace clippy clean.

Out of scope, recorded in the spec: inline PDF/video/audio rendering, streaming/Range on the blob route, a size fact, reseeding existing repos, deleting the native fallback.

---

## PR-Codex overview
This PR enhances blob rendering in the `tonk` application by introducing a declarative media view for blobs, allowing various file types to render correctly. It ensures that the blob's name is always asserted and improves the handling of blob uploads, including download functionality.

### Detailed summary
- Added `allow-downloads` permission to sandbox for blob downloads.
- Updated assertions to include the blob name in uploads.
- Introduced a new `tonk-blob/media-view` for rendering blobs, supporting different content types.
- Refactored blob image handling to use a fallback mechanism.
- Enhanced tests to verify blob name preservation and upload limits.


jackddouglas:
Live testing on a joined spot surfaced a with-forwarding gap on the space home route: the shell view mounted the content display unscoped, so `{dom.host/with}` was empty below `/space/{name}/` — `<tonk-upload>` disabled itself and the media view's URL builder bailed (entity/directory routes were unaffected, which is why the original live verification passed). Fixed in 46bc8fa6: the shell concept picks up the `xyz.tonk.site/repo` field the SW already stamps, and the shell view forwards `with="{branch}@{repo}"` exactly as the directory/entity route views do.

PR #629, opened 2026-07-22
feat(tonk-worker): hold sync drains to a quiet interval while a page is hidden

jackddouglas:
Four users were generating ~72k Cloudflare requests/day, dominated by the service worker's flat 2s sync loop running for backgrounded tabs. A hidden tab keeps its SSE subscriptions alive, so it polls all day for changes nobody is watching.

## What changed

`SyncScheduler::may_drain` — the single gate every drain path already flowed through — now enforces `max(SYNC_COOLDOWN_MS, quiet_interval(now))`:

- **Hidden page** → drains ramp geometrically from `SYNC_HIDDEN_INTERVAL_MS` (60s), doubling with continuous hidden time, capped at `SYNC_HIDDEN_MAX_MS` (1 hour). Drains land at 1, 3, 7, 15, 31, 63 min and then hourly. A tab left hidden for 24h does ~28 drains instead of 1440.
- **Visible page** → unchanged. `quiet_interval` returns `0.0`, so the gate reduces to exactly the pre-existing cooldown comparison and active collaboration keeps the 2s cadence.
- **Pending local work** → bypasses the quiet interval entirely, so a hidden tab's own un-pushed commits don't wait, however far the ramp has climbed.

`any_client_visible()` checks every window client, so "hidden" means all tonk tabs are backgrounded, not just one. Refocus posts `{type:"visibility"}` and the worker exports `onvisibility`, so regaining focus reconciles immediately rather than waiting out the interval.

Gating at the scheduler rather than the loop is what makes this work: the page POSTs `/api/sync?why=keepalive` every 10s to keep the SW alive, and every fetch schedules a debounced drain, so gating only the loop would still leave one drain per 10s.

## Notes for review

- **The ramp's origin is stamped lazily, on the first gate check after hiding — not on the visibility transition, and that is load-bearing.** The gate admits a drain on a closed `>=` against `last_drain_end`, so the ramp must still read step k−1 at the instant step k−1's interval completes. Stamping on the transition makes `hidden_since == last_drain_end`, which reads step k one instant early and silently delays every hidden drain by a full step (drains at 2/6/14/30 min instead of 1/3/7/15). `it_pins_hidden_drain_boundaries_through_may_drain` fails under that change; the reasoning is written out on `quiet_interval`.
- The dirty set that earns the bypass is kept separate from the retry set. Without that split, one repo with an unreachable relay or a non-fast-forward push would hold the bypass open forever and silently restore 2s polling against an already-erroring remote — the failure mode that would make this look shipped while fixing nothing.
- The ramp derives from wall-clock hidden time rather than a drain counter on purpose: a counter would be advanced by dirty-bypass drains, so a hidden tab actively pushing edits would climb to the cap and then be maximally stale the moment its queue emptied.
- A diagnostic logs once per hold period when the gate holds or bypasses, so the live browser check can distinguish "gate engaged" from "gate never fired".
- `test_state()` now mints a unique profile name per call. It previously shared one durable profile while three tests rename it, which made the wasm suite order-dependent — a latent bug this branch's test ordering exposed.

## Not in this PR

- **Live browser verification** has not been run. The JS glue has no automated coverage. When it is run: the ramp raises the cost of a wrong `visible=false` reading from 60s of staleness to an hour, so confirm a long-hidden tab drains immediately on refocus.
- **The companion permit cache** (dialog-db#402) reuses redeemed GET permits for 300s and roughly halves the remaining per-drain cost. Its pin bump is a separate follow-up, and `carry` pins the same tag, so bump it in step.
- The original plan's traffic arithmetic was ~2x optimistic: `publish_settled_status` calls `handle.fetch()` on every successful sync of every branch, a second Resolve beyond the pull's. The reduction ratio holds. Threading the pull's resolved revision into that classification is the best remaining lever in this repo.
- `document.visibilityState` is `"visible"` for an unfocused-but-onscreen window, so a tab parked on a second monitor still polls at 2s by design. Savings land for genuinely backgrounded tabs.

Tests: 162 wasm + 40 native passing, workspace clippy clean.

---

## PR-Codex overview
This PR implements a visibility-aware sync cadence for the service worker, introducing a quiet-interval gate that reduces sync frequency for hidden tabs and allows immediate sync upon visibility changes, thus optimizing network requests.

### Detailed summary
- Added `VisibilityState` and `WindowClient` to `Cargo.toml`.
- Implemented visibility change handling in `service_worker.js`.
- Introduced `session_nonce()` in `router.rs` for unique test profiles.
- Enhanced `SyncQueue` to manage dirty and retrying repositories separately.
- Updated `drain_sync` to prioritize dirty repositories and handle retries.
- Added tests to verify sync behavior under different conditions.
- Adjusted `SyncScheduler` to manage quiet intervals based on visibility.
- Implemented messaging from the page to the service worker on visibility changes.

> The following files were skipped due to too many changes: `rust/tonk-worker/src/worker.rs`


goblinoats:
It shouldn't be too hard to do a geometric cooldown on the SYNC_HIDDEN_INTERVAL_MS, which I like that idea, because what it means is that if I'm switching back and forth between a tab it's more likely to have a higher refresh, but if the tab is buried and not being opened then it can go through a longer reload when I come back to it. So we can potentially even increase beyond 60s to something like 300s or longer. However, if the tab has been focused in the last several hours maybe we keep the interval ~60s or something. If it's been focused in last 30min the interval is closer to 20s or something, becuase you are likely to come back to it.

jackddouglas:
yeah at some point it could be interesting to even have a durable object that pings clients when the revision cell moves so they can intelligently pull. i'm sure at a certain point that will be more economical but at this scale what we have is fine. i'm implementing exponential backoff now but even with what we have an idle tab only pings 1400 times a day which is way better than what we had

PR #635, opened 2026-07-22
chore(deps): bump dialog to tonk-2026-07-22

Gozala:
## What

Upgrades the dialog-db dependency from `tonk-2026-07-17` to `tonk-2026-07-22` and absorbs the breaking changes that tag carries. **Draft** — the core upgrade is done and verified, but one FAB-related regression is still under investigation (see "Open issue" below).

## Why

`tonk-2026-07-22` makes two large changes: the `dialog.` attribute namespace is now reserved (the artifact write path rejects any assert to it), and the binary/version-control format was redesigned (search-tree nodes, `Revision`, sync).

## Done and verified

**1. Reserved `dialog.` namespace → tonk attributes moved to `db.`**
The write path (`dialog-artifacts/tree.rs`) now rejects any `Assert`/`Replace`/`Retract` whose attribute starts with `dialog.`. tonk pervasively asserted `dialog.attribute/*`, `dialog.meta/*`, `dialog.concept/*`, `dialog.effect/*`, `dialog.name/*`, `dialog.origin/*`, `dialog.rule/*`, `dialog.view/*`. All renamed to `db.*` across Rust + the library YAML assets, including the split-domain forms (`meta_attr("dialog.X", …)`, `#[domain("dialog.X")]`, `the("dialog.X", …)`) and the four `format!` strings. Dialog-owned reserved attrs (`session`/`replica`/`branch`/`revision`/`db`) were left untouched — tonk only reads those.

**2. Dialog-injected fact renames** (these are read-only, reserved, dialog-injected — NOT renamed to `db.`, they track dialog's new names)
- `dialog.origin/{subject,profile}` → `dialog.replica/{subject,profile}` (dialog renamed its device-identity concept `Origin` → `Replica`)
- `dialog.branch/{origin,period,moment}` → `dialog.branch/{replica,revision,edition}` (the version-control redesign dropped the legacy clock fields)
- `schema::Origin` → `schema::Replica` in `tonk-worker` / `tonk-schema`

**3. Binary/version-control format**
- `Revision`/`TreeReference` moved `dialog_capability` → `dialog_artifacts`; added the dep to `tonk-worker-api`.
- `Revision` shape changed (`cause: HashSet` → `edition` + `context: Option<Context>`); rewrote `sync::classify` against `Context::includes` (now proves ahead/behind across arbitrary linear distance, not just one commit).
- `Provider<Attest>` (head-signing) added to the reactor `CommitProvider`/`PullProvider` bounds and the fixture/commit env bounds in `tonk-analyzer` (×2) and `tonk-schema`.
- Ported `dialog-reactor`'s `tree/*` inspector (`formula.rs`) to the new columnar node format (`Key` instead of `KeyBytes`, `index.links()`/`segment.keys()`+`value_at()`, EAV reconstructed from the key via `Artifact::from_key_datum_placeholder`, manifest-aware `Geometric::rank`).

**Verification** (live, against a running local instance):
- CORE seeds clean; `blank:` resolves the space subject via `dialog.replica/subject`; `tonk/repository` resolves the renamed name; `concept:origin`/`space:` resolve; pull/push succeed; spaces render.
- Native + wasm builds clean for all bundle crates; native test suites pass.
- Regression tests added: `it_binds_the_dialog_injected_replica_subject` and `it_persists_a_space_rename` (both wasm/service-worker tests). The rename test **passes** through the `evaluate`/rule path — the rename rule itself works.

**Gotchas that bit us (worth remembering):**
- Several breaks were `#[cfg(wasm32)]`-gated, so a native `cargo check --workspace` passed while the trunk (wasm) build failed. Always `cargo check --target wasm32-unknown-unknown -p tonk-worker -p tonk-guest -p tonk-ui` after a dialog-facing change.
- Trunk silently serves the last-good `dist/` when a build fails, and leaves stale hashed snippet dirs, which caused a `WebAssembly.instantiate: Import … inline0.js` SW crash. A clean `rm -rf rust/tonk-ui/dist` + rebuild fixes it.
- The `db.rule/*` rename **converges** with dialog's native deductive-rule storage (not a collision). Heads-up: dialog PR #405 will rename `db.rule` → `dialog.rule` (moving it under the reserved namespace) — so tonk's `db.rule/*` will need to become `dialog.rule/*` (read-only) once #405 lands.

## Open issue — FAB regression (still debugging)

On this branch, in-space the FAB dock chip shows **"Untitled"** (not the renamed name), the sync-status chip reads **offline**, the **share** button spins, and renaming shows **no visible transactions**. These work on `origin/staging`.

**Confirmed NOT the cause** (all verified identical to staging / correct on this branch):
- The data layer: `tonk/repository name`, `blank` subject, the FAB's exact `xyz.tonk.repo/name` inline query all resolve correctly via direct HTTP.
- The FAB code, the routes, and the `xyz.tonk.site/repo`-vs-`space` naming are unchanged from staging (my commit doesn't touch `tonk-fab` at all).
- The served bundle is fresh; there are no app-originated failing requests.

**Localized to:** the sealed guest renders `<tonk-site with="main@">` (empty repo) → `route_from` rejects the malformed `with` → the subscription's host never writes `detail.subscription` → the FAB/status subscriptions fail → "Untitled" + no live updates. The divergence from staging: querying `workspace/shell` on the space **content branch**, `xyz.tonk.site/repo` is **populated on staging but empty on this branch**. `stamp_site_on` stamps `("repo", repo)` explicitly, and is gated by `origin_entity` (the `Origin`→`Replica` query I changed). The current hypothesis is that the space-mode site stamp is landing with an empty `repo` (or being skipped), leaving only the profile-mode (empty-repo) stamp for the guest to read.

**Diagnostic commit** (`debug: introspection …`, to be reverted before merge) adds logging to `dispatch` (arriving transient attributes + matched-handler count) and `stamp_site_on` (`repo`/`branch`/`profile`/route it stamps). Next step: reload with a fresh worker build and read the **SW console** to see exactly what `repo`/`profile` the guest's content-branch site stamp carries, and whether the FAB rename transient reaches `dispatch` at all (the "no transactions" symptom suggests it may not).

---

## PR-Codex overview
This PR focuses on refactoring the naming conventions and attribute references across the codebase, primarily changing instances of `dialog` to `db` in various files, which likely indicates a shift in the underlying architecture or data model.

### Detailed summary
- Changed `dialog.name/referent` to `db.name/referent` in multiple files.
- Updated attribute references from `dialog.meta` to `db.meta`.
- Replaced `dialog.effect` with `db.effect` in effect-related queries.
- Modified `dialog.capability` to `dialog.artifacts` in relevant imports.
- Adjusted YAML files to reflect the new naming conventions for subjects.
- Refactored function signatures and internal logic to accommodate the new attribute names.

> The following files were skipped due to too many changes: `rust/tonk-schema/src/resolution.rs`, `rust/tonk-worker/src/router/repository.rs`, `rust/tonk-schema/src/transact.rs`, `rust/tonk-schema/src/rule_query.rs`, `Cargo.toml`, `rust/tonk-core/src/effect.rs`, `rust/tonk-tree/src/web.rs`, `rust/tonk-analyzer/src/analyzer/declaration.rs`, `rust/tonk-analyzer/src/analyzer.rs`, `rust/tonk-worker-api/src/sync.rs`, `rust/tonk-evaluator/src/evaluate.rs`, `Cargo.lock`, `rust/tonk-schema/src/rule.rs`, `rust/tonk-tree/src/key.rs`, `rust/dialog-reactor/src/formula.rs`, `rust/tonk-schema/src/concept.rs`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -919,7 +919,7 @@ dependencies = [
 [[package]]
 name = "dialog-artifacts"
 version = "0.1.0"
-source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-17#2e18c18d57425aa7d39321ceda6a5816c9635881"
+source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-22#534bf48cce31346b72635c9deb2de948382240f5"
 dependencies = [
  "arrayref",
  "async-stream",
@@ -942,10 +942,13 @@ dependencies = [
  "rkyv",
  "serde",
  "serde-big-array",
+ "serde_bytes",
+ "serde_ipld_dagcbor",
  "serde_json",
  "static_assertions",
  "thiserror 2.0.18",
  "tokio",
+ "tracing",
  "url",
  "wasm-bindgen",
  "wasm-bindgen-futures",
@@ -954,16 +957,21 @@ dependencies = [
 [[package]]
 name = "dialog-capability"
 version = "0.1.0"
-source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-17#2e18c18d57425aa7d39321ceda6a5816c9635881"
+source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-22#534bf48cce31346b72635c9deb2de948382240f5"
 dependencies = [
  "async-trait",
  "base58",
+ "blake3",
  "convert_case",
  "dialog-common",
  "dialog-macros",
  "dialog-varsig",
+ "ed25519-dalek",
+ "rkyv",
  "serde",
+ "serde-big-array",
  "serde_bytes",
+ "serde_ipld_dagcbor",
  "signature",
  "thiserror 2.0.18",
  "url",
@@ -972,7 +980,7 @@ dependencies = [
 [[package]]
 name = "dialog-common"
 version = "0.1.0"
-source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-17#2e18c18d57425aa7d39321ceda6a5816c9635881"
+source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-22#534bf48cce31346b72635c9deb2de948382240f5"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -1000,7 +1008,7 @@ dependencies = [
 [[package]]
 name = "dialog-credentials"
 version = "0.1.0"
-source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-17#2e18c18d57425aa7d39321ceda6a5816c9635881"
+source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-22#534bf48cce31346b72635c9deb2de948382240f5"
 dependencies = [
  "async-trait",
  "base58",
@@ -1021,7 +1029,7 @@ dependencies = [
 [[package]]
 name = "dialog-csv"
 version = "0.1.0"
-source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-17#2e18c18d57425aa7d39321ceda6a5816c9635881"
+source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-22#534bf48cce31346b72635c9deb2de948382240f5"
 dependencies = [
  "async-trait",
  "base58",
@@ -1036,7 +1044,7 @@ dependencies = [
 [[package]]
 name = "dialog-effects"
 version = "0.1.0"
-source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-17#2e18c18d57425aa7d39321ceda6a5816c9635881"
+source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-22#534bf48cce31346b72635c9deb2de948382240f5"
 dependencies = [
  "async-trait",
  "base58",
@@ -1051,7 +1059,7 @@ dependencies = [
 [[package]]
 name = "dialog-macros"
 version = "0.1.0"
-source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-17#2e18c18d57425aa7d39321ceda6a5816c9635881"
+source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-22#534bf48cce31346b72635c9deb2de948382240f5"
 dependencies = [
  "blake3",
  "convert_case",
@@ -1063,7 +1071,7 @@ dependencies = [
 [[package]]
 name = "dialog-network"
 version = "0.1.0"
-source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-17#2e18c18d57425aa7d39321ceda6a5816c9635881"
+source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-22#534bf48cce31346b72635c9deb2de948382240f5"
 dependencies = [
  "async-trait",
  "dialog-capability",
@@ -1078,7 +1086,7 @@ dependencies = [
 [[package]]
 name = "dialog-operator"
 version = "0.1.0"
-source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-17#2e18c18d57425aa7d39321ceda6a5816c9635881"
+source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-22#534bf48cce31346b72635c9deb2de948382240f5"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -1112,7 +1120,7 @@ dependencies = [
 [[package]]
 name = "dialog-query"
 version = "0.1.0"
-source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-17#2e18c18d57425aa7d39321ceda6a5816c9635881"
+source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-07-22#534bf48cce31346b72635c9deb2de948382240f5"
 dependencies = [
  "async-stream",
  "async-trait",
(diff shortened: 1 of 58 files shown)


Gozala:
### Update — FAB root cause pinned to unsubstituted `{id}`

Instrumented `<tonk-fab>` (`[fab-dbg]` in the `debug:` commit) to log the `space` attribute it renders with (guest logs bubble to the page console). On this branch, in the sealed space frame:

```
[fab-dbg] inject_children space="{id}"     ← LITERAL, unsubstituted placeholder
tonk-site: malformed with="main@": empty repo after '@'
[fab-dbg] inject_children space=""          ← a second FAB, empty
ui-sync-status: subscribe failed: host did not write detail.subscription
```

**The FAB receives the literal string `"{id}"`, not the space DID.** So `<tonk-fab space="{id}">` in the space-view (`tonk:space/chrome`, route `/space/{id}`) is rendered **without substituting the `{id}` binding**. Everything downstream follows: `space_with("{id}")` → a malformed `with`, the subscription's host rejects it → `detail.subscription` never written → "Untitled" + no live updates + the sync chip offline. The "no transactions" symptom is the same cause — the rename input's context is a dead placeholder.

`{id}` = `xyz.tonk.site/id`, a **required** field of the `space-chrome` concept. When queried directly (profile branch) it resolves populated, but at **render time in the guest** the concept resolves 0 rows, so the `{id}` binding stays literal. On `origin/staging` the same view substitutes `{id}` correctly.

None of the involved code (`tonk-fab`, the routes, `tonk-render`/`tonk-display` substitution, the `match_route` param capture) is touched by this branch — so the regression is a *downstream* effect of the dialog fact renames on how `space-chrome` resolves at render time in the sealed guest (a stamp/render race, a failed subscription, or a transitive dependency on a renamed dialog-injected fact).

**Next step:** determine why `space-chrome` (required `id`) resolves empty at guest render time when its `xyz.tonk.site/id` fact is stamped — instrument the guest's `space-chrome` subscription/resolution, or check whether the render races the site stamp.

Gozala:
### Refinement — `space-chrome` resolves empty at guest render

Same reload, two FAB renders, no later one (so it is NOT a race that eventually resolves):

```
[fab-dbg] inject_children space="{id}"    ← FAB #1: LITERAL placeholder (binding never ran)
tonk-site: malformed with="main@"          ← same template's {id} substituted to EMPTY
[fab-dbg] inject_children space=""          ← FAB #2: empty
```

Two things stand out in the **same** `space-view` template (`tonk:space/view`, model `tonk:space/chrome`):
- `<tonk-site with="main@{id}">` → `main@` — here `{id}` substituted to the **empty string**.
- `<tonk-fab space="{id}">` → `{id}` — here it stayed the **literal placeholder**.

So `space-chrome.id` (`xyz.tonk.site/id`, a **required** field) resolves to empty/absent at guest render time, even though a direct profile-branch query returns it populated (the DID). Because `id` is required-cardinality-one, an empty `id` means `space-chrome` yields no usable row, and the template's `{id}` bindings degrade (empty for the substituted one, literal for the FAB's). Everything downstream — the malformed `with`, the failed `detail.subscription`, "Untitled", no live updates, the offline sync chip, the dead rename — follows from `space-chrome` not resolving `id`.

This is the crisp reproduction to hand off. The regression is that `space-chrome`'s `xyz.tonk.site/id` binding does not resolve during the sealed-guest render on this branch (it does on staging), despite the fact being stamped. Likely candidates: the site re-stamp's `retain_overlay_entities` wipe racing the guest render, or the guest's `space-chrome` subscription transitively depending on a renamed dialog-injected fact. The `[fab-dbg]`/`[dispatch-dbg]`/`[stamp-dbg]` instrumentation (in the `debug:` commit) is in place to continue.

Gozala:
### FAB regression FIXED ✅

Root-caused and fixed. In the sealed space frame the shell stamps the FAB with `space="{id}"` and `<tonk-site with="main@{id}">` **before** the site's `id` field resolves, then rewrites the attributes once it lands. Three gaps kept the FAB stuck on "Untitled" with an offline sync chip and a dead rename:

1. **Multi-segment binding substituted an absent field to nothing.** `with="main@{id}"` with `id` absent rendered `with="main@"` — a value with no `{…}` left, so `<tonk-site>` took it for a resolved-but-*malformed* location and errored permanently instead of waiting. Fix: hold off writing partially-resolved multi-segment attribute bindings until the field lands (mirrors the existing lone-`{field}` absent handling). *(`tonk-display`)*

2. **`<tonk-fab>` didn't observe `space`**, so the bar rendered once with the placeholder and never updated. Fix: observe `space` and propagate the resolved DID to the space-scoped children. *(`tonk-fab`)*

3. **`<ui-space-name>` / `<ui-sync-status>` had no `attribute_changed_callback`**, so a `space`/`with` that resolved after mount never drove a re-subscribe. Fix: add the callback so they connect once the real value arrives. *(`tonk-fab`, `tonk-workspace`)*

Verified live: the FAB shows the real space name ("Debug"), the sync chip is healthy (not offline), the rename persists, and the console is clean — no `malformed with="main@"`, no `subscribe failed`.

The `debug:` introspection commit has been dropped; its findings are folded into the `fix(fab):` commit. The branch is now two commits: the dialog upgrade + the FAB fix.

Gozala:
### Added — `/diagnose` tree inspector adapted to the M3 key format

The search-tree inspector still decoded keys as the old fixed 162-byte layout, so every variable-length M3 key fell back to one undifferentiated hex blob and node separators/bounds read as raw hex.

Now the keys are decoded **on the worker** (`dialog-reactor`'s `key_parts`, which has the real dialog `Key`/`KeyView` API): each `tree/*` conclusion carries a structured, self-describing `parts` list, and the `tonk-tree` UI renders those as colored, labeled chips.

- **Textual components** — the entity as its `did:key:…` URI, the attribute as its `db.x/y` name, the value typed and textualized (strings quoted, numbers signed, records/bytes size-labeled). Verified live: e.g. `entity:concept:5ks… · attribute:db.meta/description · vtype:String · value:"An inspector tile body…"`.
- **Separators / bounds** — decoded per the child link's ordering, front-coding pivot still computed on the raw bytes.
- **Glyphs for structural bytes** — a NUL/control byte renders as `␀`/`·` so structure is visible instead of an invisible gap.
- **History & blob orderings** — previously opaque, now decoded: history keys surface the version prefix (`origin`, `edition`) then the fact fields; blob keys show the content-addressed hash.
- **`dialog.diagnose/name` → `db.diagnose/name`** — the diagnose concept's one field was still on the reserved namespace (my earlier sweep missed `diagnose`).

Also folded out the temporary `[dispatch-dbg]`/`[stamp-dbg]` introspection that had leaked into the FAB-fix commit — the branch is now three clean commits (dialog upgrade, FAB fix, diagnose), fmt + clippy clean, native + wasm building.

PR #647, opened 2026-07-27
feat(cli): scope active spots to working directories

jackddouglas:
## Summary

Replace the machine-global active spot with persistent directory bindings in the central spot registry.

```text
--spot > TONK_SPOT > nearest bound ancestor of cwd
```

There is no global fallback. A directory is only a canonical lookup key; spot data remains in its central site directory.

```sh
tonk use garden       # use garden in this directory and descendants
tonk spot unbind      # remove this directory's exact binding
tonk spot list        # show registered spots, bindings, and what is active here
tonk status           # show the active spot, resolution source, and sync state
```

## Behavior

- `tonk use <name>` always binds the current directory; the redundant `--here` flag is removed.
- `tonk spot new` and `tonk join` bind their invocation directory on success, preserving create-and-activate behavior without shared global state.
- Nested bindings override parent bindings. `--spot` and `TONK_SPOT` remain explicit invocation and process overrides.
- `tonk spot detach` becomes `tonk spot unbind`, which matches an exact path and names a bound ancestor rather than removing it implicitly.
- `tonk spot list` reports `active here` and lists a `directories:` block.
- Spot-scoped failures append the active spot, resolution source, and central site to stderr so humans and agents can see which context failed.

## Registry migration

`spots.json` now stores a top-level `bindings` map from canonical directory path to spot name. Registries using the unreleased `attachments` name deserialize into `bindings`.

The old `current` field is accepted only as migration input. Resolution never consults it, and the next registry write drops it. Unknown fields continue to survive round trips.

## Why `use`

`use` stays as the intent-level verb: “use this spot here.” `worktree` would imply a copied working tree, while `mount` would imply filesystem exposure; Tonk does neither. The implementation and documentation call the underlying pointer a directory binding.

This borrows the useful part of Git worktrees—a durable central-repository-to-directory association—without copying Tonk data into each working directory or reviving cwd-based `.tonk/` discovery.

## Verification

- `cargo test -p tonk-cli`: 276 passed, 1 pre-existing ignored
- `cargo clippy -p tonk-cli --all-targets -- -D warnings`
- `cargo fmt --all -- --check`
- `cargo check -p tonk-cli --all-targets`
- `git diff --check`

The revised design and implementation notes live under `docs/superpowers/`.


goblinoats:
Thought about this a bit and I have a competing idea I would like to propose

PR #648, opened 2026-07-27
feat(tonk-cli): resolve spots from TONK_SPOT only, add `tonk spot enter`

goblinoats:
## Why

Spot selection was a machine-wide `current` in `spots.json`: any terminal could rewrite it, none of them displayed it, and every bare command fell back to it. Two terminals could not hold different spots without repeating `--spot` on every command, and `tonk use` in one window silently repointed the others.

## What changes

One selector, no default:

```
TONK_SPOT is set  ->  that spot
otherwise         ->  error naming the ways forward
```

A **reference** is a registered name (`garden`) or a path to a site directory (`~/proj/.tonk`) — anything that isn't a legal name slug is read as a path, so a mistyped path reports `no site directory at /abs/path` rather than `unknown spot`.

Two ways to get one, both of them things you typed:

- **`tonk spot enter [ref]`** runs `$SHELL` with `TONK_SPOT` exported, so a terminal keeps its spot and two terminals can hold different ones. The prompt gains a `(tonk: <name>)` segment. Bare `tonk spot enter` starts from `./.tonk`.
- **`TONK_SPOT=garden tonk …`** sets it for one process — what agents and CI use, since they can't live in a subshell.

Every command prints the spot it used on stderr (stdout stays data-only, so piping and `--json` are unaffected).

## Notable details

**Prompt decoration is shell-specific by necessity.** A prompt can't be passed through the environment — startup files run after we spawn and assign `PS1` themselves. bash goes via `--rcfile`, zsh via `ZDOTDIR`, fish via `-C`. The generated files source the user's own config *first*, so a bug in this layer can't cost someone their shell setup, and zsh's hands `ZDOTDIR` back so nothing launched from that session inherits the redirect. Unknown shells launch plain with a hint; `TONK_NO_PROMPT=1` opts out. Prompt frameworks (starship, powerlevel10k) rebuild the prompt each redraw and will overwrite the segment — which is why the stderr announce exists as the actual guarantee.

The segment reads `TONK_SPOT` at redraw rather than baking the name, so it follows an `export` and empties on `unset`. zsh only substitutes parameters in prompts under `PROMPT_SUBST`, so it gets the dynamic form only when that's already set — turning it on under someone else's theme could change how their prompt evaluates.

**Inheritance warning.** A human enters a spot and launches an agent from that shell; the agent inherits `TONK_SPOT`, never sets it, and quietly succeeds against whatever the human was on. This can't be detected directly — `TONK_SPOT=x tonk …` and `export TONK_SPOT=x` reach the process byte-identical — so `spot enter` also exports `TONK_SPOT_SESSION`, and comparing the two answers it. Warns when they agree *and* stderr isn't a terminal. A human in their own entered shell is never nagged, and `tonk query | grep` keeps stderr on the tty. It stays a warning because an agent launched inside an entered shell may well be meant to work there.

Known limit, documented rather than papered over: inheritance from an `export TONK_SPOT` in a shell profile is invisible to this, since there's no session to compare against.

## Breaking

- **`tonk use` is removed** — errors pointing at `tonk spot enter`.
- **`--spot` is removed** — it was a second spelling of `TONK_SPOT` that could disagree with it. Use `TONK_SPOT=<ref> tonk <cmd>`.
- **The working directory is no longer consulted.** A `.tonk` beside you is not a selection; only `tonk spot enter` starts from it.
- **`tonk spot new` and `tonk join` no longer select** what they create — they print the enter line instead.
- A `current` key left by an older tonk parses fine and is dropped on the next write.

Anyone with a habit of bare `tonk query` will need `TONK_SPOT` set; a personal default belongs in your own shell profile rather than in tonk's state directory.

## Testing

16 test binaries green, clippy and fmt clean. Beyond unit and CLI-level tests, the shell paths were verified against real bash and zsh under a pty: prompt decoration renders, the segment tracks the variable, and the inheritance warning fires for captured output but not for an interactive user.

---

## PR-Codex overview
This PR focuses on refining the `tonk` command-line tool's spot management, emphasizing a single source for spot resolution and enhancing documentation for clarity.

### Detailed summary
- Added a new module `shell` for managing subshells with spot information.
- Updated documentation to clarify spot resolution processes.
- Removed the `tonk use` command, replacing it with `tonk spot enter`.
- Enhanced error messages for spot selection failures.
- Revised tests to align with the new spot management logic.

> The following files were skipped due to too many changes: `rust/tonk-cli/src/spot.rs`, `rust/tonk-cli/src/bin/tonk.rs`


jackddouglas:
i like putting the spot on `stderr` and `spot enter` is good for a long-lived terminal session but how does it work with agents? in my tests claude code/codex inherits env vars set from the launching shell but drop env vars set in any command run by the agent inside the harness so it'd still need to specify the spot each invocation if you don't set it in the launching shell beforehand.

i could get behind removing the global `tonk use` but i think the env var is the most fragile mechanism so it definitely shouldn't be the default. it's not transparent or durable.

`TONK_SPOT=x tonk …` is worse than `--spot` imo and still requires you to pass it each bash command.

also if you enter a spot and launch an agent from that shell it'll inherit `TONK_SPOT` and `TONK_SPOT_SESSION` and if they agree and stderr isn't a tty then it warns on every command that they were inherited, even if it was successful, which is noisy for the desired usage. makes sense to give the agent context about what spot is selected and why on errors, but not on every command.

either way i wouldn't get rid of the `--spot` flag because it's useful for scripts, composability, and anywhere you can't/don't want to control env vars

jackddouglas:
wtf lmfao

jackddouglas:
what don't you like about `tonk use` setting an alias to a working directory? if we remove the global `tonk use` and have it strictly alias dirs then you get much better transparency and it's more predictable than env vars. also much closer to the git worktree model people understand well

PR #654, opened 2026-07-30
ci(cli-npm): allow manual npm publish runs

jackddouglas:
## What

`.github/workflows/cli-npm.yml` only triggered on a `v*` tag push, so pushing a tag was the only way to publish `@tonk/cli`. A publish that failed after the tag already existed could not be retried without cutting a new tag.

This adds `workflow_dispatch` with an optional `version` input:

- **Empty input** — publishes whatever `[workspace.package] version` says on the ref you select.
- **Supplied value** (`0.6.3` or `v0.6.3`) — only asserts it; a mismatch against the ref's Cargo workspace version still hard-fails.

The Cargo workspace version stays the single source of truth, and dist-tag routing is unchanged (hyphenated -> `next`, otherwise `latest`).

## Caveat

npm refuses to republish an existing version, so dispatch only rescues a partial failure where the failing package never landed. Otherwise a version bump is still required.

## Verification

- `actionlint .github/workflows/cli-npm.yml` — exit 0.
- Extracted the resolve script and ran it against the real `Cargo.toml` (0.6.3):
  - dispatch / empty input -> `0.6.3`, `latest`
  - dispatch / `v0.6.3` -> `0.6.3`, `latest`
  - dispatch / `9.9.9` -> fails with a version-mismatch error
  - tag push `v0.6.3` -> `0.6.3`, `latest`
  - tag push `v0.1.0` -> fails with a version-mismatch error

---

## PR-Codex overview
This PR enhances the GitHub Actions workflow for publishing npm packages by adding a manual trigger option and improving version validation logic.

### Detailed summary
- Added `workflow_dispatch` to allow manual execution of the workflow.
- Introduced an `inputs.version` parameter for optional version specification.
- Improved version validation logic to handle manual and tag-based triggers.
- Added a verification step for npm authentication before publishing.
- Refactored the publishing process to skip already published packages.

### DIFF
--- a/.github/workflows/cli-npm.yml
+++ b/.github/workflows/cli-npm.yml
@@ -4,10 +4,21 @@ name: 'CLI npm'
 # and publishes @tonk/cli (plus its platform packages) to npm. The
 # version comes from the tag and must match the Cargo workspace version;
 # nothing publishes on a normal push.
+#
+# Can also be run manually against any ref (`workflow_dispatch`), for
+# re-running a failed publish or shipping without cutting a tag. The
+# version still has to match the Cargo workspace version of the ref you
+# pick, so there is only ever one source of truth.
 on:
   push:
     tags:
       - 'v*'
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version to publish. Leave empty to use the ref Cargo workspace version.'
+        required: false
+        type: string
 
 concurrency:
   group: cli-npm-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


jackddouglas:
Pushed a second commit (649d94bf3) after digging into why the tag path never worked.

**Every `CLI npm` run has failed** — `v0.4.0`, `v0.5.0`, and one dispatch. The `v0.5.0` failure was `npm error 404 Not Found - PUT https://registry.npmjs.org/@tonk%2fcli-darwin-arm64`, and that package already existed at `0.4.0` at the time, so it was an auth failure wearing a 404. npm `@tonk/cli@0.4.0` was published by hand ~40 min after its CI run failed.

Two guards, both aimed at that:

- **`npm whoami` preflight** — a bad or under-scoped token now fails in seconds with an auth error instead of a 404 on PUT that reads like a missing package.
- **Skip already-published versions** — the publish step ran three bare `npm publish` calls under `set -e`. If the second failed, the first stayed published forever, and since npm never permits republishing a version, the retry this PR adds would have been dead on arrival and the version number burned. The loop now skips any package already on the registry, so a retry finishes the remainder.

Verified: `actionlint` exit 0, and the name resolution + skip predicate run against the live registry (`0.4.0` -> SKIP for all three, `0.6.3` -> PUBLISH for all three).

Note this workflow is dispatchable only once it is on `staging`, since `workflow_dispatch` has to exist on the default branch.

PR #700, opened 2026-08-07
Passkey fan-out: multiple credentials per account

jackddouglas:
One account, several passkeys, none of them the account. The full design is
in `plan/passkey-fanout.md`; this is the specification of what the branch
makes true.

## Identity

An account's subject is an Ed25519 keypair generated during genesis and
destroyed before genesis returns. Nothing holds it afterwards.

A passkey's PRF output derives a **credential key** — the same derivation as
today, `tonk/root-key/v1`, unchanged. What changes is its meaning: it is one
credential among peers rather than the account itself.

Every credential reaches the subject through a chain of subject-open,
command-open delegations. A device reaches it through a credential. Depth is
not fixed and carries no meaning: `root → C1 → device`,
`root → recovery → C2 → device` and longer are equally valid.

## Genesis

The subject signs, in this order, and is then dropped:

1. the account repository descriptor (unchanged v1 format),
2. `root → C1`, to the first passkey's credential key,
3. `root → recovery`, to the account service's anchor DID.

**Anchor invariant.** Because the subject is destroyed, every direct
`root → x` delegation that will ever exist is minted here. Afterwards a new
anchor is reachable only by an existing credential delegating one. An account
that leaves genesis without a recovery anchor can never be given one except
through a live credential.

## Enrollment

A credential is enrolled by receiving a subject-open delegation. Two paths
produce one, and a credential may hold chains from both:

- **Anchor path**, the flow a person walks. The service verifies a one-time
  code sent to the account address, mints `recovery → Cn` under its genesis
  proof, publishes it, and mails the account address that it did. Requires
  nothing but the new passkey — no second device.
- **Sibling path**, taken opportunistically when a credential is reachable
  anyway. `Ca → Cn`, minted locally, no service involved. Never a step the
  person is asked to perform.

Anchor issuance is gated on the code **even when the caller already holds a
valid chain**. An anchor chain outlives revocation of whatever enrolled it,
so gating on "holds a chain" would let a briefly compromised passkey mint
itself a successor that survives being revoked.

## Chain store

A device that has just derived `Cn` holds nothing else — not the subject, not
a delegation, and not the account repository, whose sync authority is the
thing it is looking for. It asks what has been addressed to the one key it
has.

- `POST /enrollments` publishes a chain. Unauthenticated: the bytes are
  signed and self-authenticating, so publishing confers nothing.
- `GET /enrollments/{credential_did}` returns every chain addressed to that
  key. Unauthenticated: knowing a DID is not authority, and a delegation is
  useless without the matching private key.

The store can withhold an entry. It cannot forge, alter, or grant by holding
one. Every entry also belongs in the account repository, so the user or a
competing provider can serve the same bytes.

Publication additionally requires that this service hosts the account the
chain runs from. That is anti-abuse — the chain itself already settles who
may enrol — and keeps the bucket from being open storage.

## Revocation

Verification is unchanged in mechanism: a valid, unrevoked chain from the
subject to the invoker, any one of several. What is added is an authority
class, because neither existing class covers a peer withdrawing a link it
does not depend on — Safari holding only `root → recovery → safari` has no
standing over `root → chrome` under either.

- **Peer** — the signer presents its own currently-valid chain from the same
  subject-open authority the target answers to.

Gated on the witnessed path being subject-open, so space invites, which are
subject-specific, keep the narrower classes and the two delegation shapes
stay distinct.

## Positions this takes

**The recovery anchor is a delegation, not an authority.** It is a real
keypair whose standing comes from one visible link the user can cut. This is
deliberately not storacha's `ucan/attest`, which roots the equivalent power
in the service's own DID, where no verifier can decline it and no user can
revoke it. They need that indirection because `did:mailto` cannot sign; an
anchor holding a key does not.

**Email control becomes account control.** That is the cost of the anchor
path being the flow, and it is stated rather than mitigated away. The
containments are that the link is revocable by any peer and that every
issuance is announced to the account address.

**Losing the email and every passkey loses the account.** There is no
service-side key beyond the anchor the user already sees in their own chain.

## Not in this branch

The enrollment log and the revocation walk that reads it; the seniority rule
for mutual revocation; browser wiring for any of these ceremonies; v1
accounts publishing `subject → recovery` at sign-in.

Two decisions are deliberately unmade and recorded in the plan: whether
`Peer` is admissible at the account service's device-revocation gate, and
whether the enrollment log is an authority source for account-level acts.
The second blocks the first.

Separately, and predating this work: revocation artifacts are screened
globally by target CID, so anyone holding a delegation's bytes can prepend a
link they mint and qualify as `PathIssuer` over it. Untouched here and
described in the plan's status section.

## Deployment

`RECOVERY_ANCHOR_SEED` is a per-environment worker secret, 32 bytes hex, set
with `wrangler secret put` — deliberately absent from `wrangler.account.toml`
so production, staging and preview hold different anchors. Without it every
other route behaves normally and `GET /` omits `recoveryAnchor`. No new
bindings, buckets or migrations: enrollment objects share the existing
`CHAINS` bucket under their own prefixes.

https://claude.ai/code/session_01CAfmVJYm3tHXZmYToJQqnE

---

## PR-Codex overview
This PR introduces enhancements to the `tonk-account-service`, focusing on enrollment functionalities, error handling, and the integration of a new `zeroize` dependency for secure key management. It also refines email handling and adds new enrollment routes.

### Detailed summary
- Added `pub mod enrollment` to `src/core.rs`.
- Introduced `zeroize` dependency in `Cargo.toml`.
- Enhanced email handling with `send_enrollment_notice`.
- Added new routes for enrollment: `handle_publish`, `handle_confirm`, `handle_claim`.
- Implemented `R2EnrollmentStore` for managing enrollments.
- Refined error handling and response structures.
- Updated `build_anchor` and `build_enrollments` functions.
- Improved tests for enrollment and delegation functionalities.

> The following files were skipped due to too many changes: `rust/tonk-account-service/src/core/enrollment.rs`, `rust/tonk-account-service/tests/service.rs`, `rust/tonk-account-service/src/enrollments.rs`, `plan/passkey-fanout.md`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4524,6 +4524,7 @@ dependencies = [
  "wasm-bindgen-test",
  "worker",
  "worker-macros",
+ "zeroize",
 ]
 
 [[package]]
--- a/plan/passkey-fanout.md
+++ b/plan/passkey-fanout.md
@@ -108,6 +108,43 @@ Smallest stable slice first; each step is independently shippable and useful.
 4. **Anchor normalization and email-gated recovery** in the account service.
 5. **v1 opt-in fan-out** (`subject → recovery` minting at sign-in) and the revocation walk UX.
 
+## Status
+
+Landed on `feat/passkey-fanout`:
+
+- **Step 1.** `RevocationAuthority::Peer` in `tonk-identity/src/revocation.rs`, plus
+  `mint_peer_revocation`. Gated on the witnessed path being subject-open, so
+  space invites keep the narrower classes.
+- **Step 2, enabling half.** `tonk-identity/src/credential.rs` mints the
+  `credential → peer` enrollment link and composes chains under it;
+  `delegation::validate_account_grant` replaces the five separate
+  "exactly one proof" checks in the account service, worker and CLI, so a
+  device may present `root → credential → device` at any depth. The
+  registry now records the device's own hop rather than the chain root —
+  identical for a one-hop grant, so no migration.
+
+Not started: the browser enrollment ceremony and wherever the enrollment link
+is served from (a device holding only the second passkey has no way to obtain
+`root → C2` today), and steps 3–5.
+
+Two decisions the implementation surfaced, both left open deliberately:
+
+- **Peer at the account service's device-revocation gate.** `revoke_device`
+  admits only `Delegated` (a device revoking itself) and `PathIssuer` (the
+  account root). Admitting `Peer` would also let any *device* revoke any other
+  device, because the service cannot tell a credential from a device — both
+  hold subject-open chains from the root. That drops the current requirement
+  that cross-device revocation re-derive the root from the passkey. Wants a
+  decision before it is written.
+- **Revocation artifacts are screened globally by target CID.** An artifact is
+  authorized relative to its witnessed path, but `tonk-access-service` and the
+  R2 key record only the target CID. Anyone who holds a delegation's bytes can
+  mint `attacker → issuer`, prepend it to make the target sit at index 1, and
+  qualify as `PathIssuer` — which globally revokes it. Reachable today by any
+  invitee against their inviter's link. Predates this work and is untouched by
+  it; subject-scoping the screen does not fix it, because subject-open grants
+  are meant to apply across subjects.
+
 ## Open questions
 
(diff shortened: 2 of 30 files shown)


jackddouglas:
Superseded by #701, which is the spec on its own. The implementation stays on `feat/passkey-fanout` and can be reopened once the spec settles.

PR #701, opened 2026-08-07
Spec: passkey fan-out

jackddouglas:
## PR-Codex overview
This PR introduces the concept of `passkey fan-out`, detailing how multiple passkeys can be associated with a single account without binding it to a specific passkey. It outlines the model, genesis, enrollment, verification, revocation, recovery, and storage of these passkeys.

### Detailed summary
- Added a new section on `Passkey fan-out (V1)`.
- Defined key concepts: `Account subject`, `Credential key`, `Device key`, `Anchor`, and `Recovery anchor`.
- Described the `Genesis` process for account creation.
- Explained `Enrollment` paths for credentials.
- Outlined `Chain store` operations for devices.
- Provided details on `Verification` and `Revocation` processes.
- Discussed `Recovery` methods for lost credentials.
- Explained `Storage and portability` of accounts.
- Addressed `Existing accounts` and their compatibility.
- Included a `Rationale` for the design choices.
- Listed `Deferred` items and `Open questions` for future consideration.

### DIFF
--- a/docs/passkey-fanout.md
+++ b/docs/passkey-fanout.md
@@ -1,0 +1,274 @@
+# Passkey fan-out (V1)
+
+One account, several passkeys, none of them the account.
+
+A passkey is bound to one authenticator, so an account whose identity *is* a
+passkey reaches only the platforms that authenticator reaches. Chrome on a Mac
+and Safari on an iPhone are routinely not the same set. This specification
+breaks the identity `account subject = HKDF(passkey PRF)` without coupling the
+account to any single passkey, without hiding authority state from the user,
+and without binding the account to Tonk's account service any harder than it
+is bound today.
+
+## Model
+
+- **Account subject** — an Ed25519 keypair generated during genesis and
+  destroyed before genesis completes. Its DID names the account forever.
+  Nothing holds the key afterwards.
+- **Credential key** — the key derived from one passkey's PRF output. The
+  derivation (`tonk/root-key/v1`) is unchanged; its meaning changes from "the
+  account" to "one credential among peers".
+- **Device key** — a per-device key that acts through a credential.
+- **Anchor** — the audience of a direct `subject → x` delegation minted at
+  genesis. Initially the first credential and the recovery anchor.
+- **Recovery anchor** — a provider's own keypair, holding a direct
+  subject-open delegation. A competing provider added later is another
+  anchor, reached through a credential rather than through the subject.
+
+Account authority is carried by **subject-open, audience-specific, command-open**
+delegations. Space invites are subject-specific, and the two shapes must not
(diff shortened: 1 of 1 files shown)


docs/passkey-fanout.md
```diff
@@ -0,0 +1,274 @@
...
+must be able to confer full account authority on a replacement credential. The
+containment is structural rather than syntactic: the delegation is one visible
+link in the user's own chain, revocable by any peer.
+
+**Anchor invariant.** Because the subject is destroyed, every direct
+`subject → x` delegation that will ever exist is minted here. Afterwards a new
+anchor is reachable only by an existing credential delegating one. An account
+that leaves genesis without a recovery anchor can never be given one except
+through a live credential, and if every credential is lost first, the account
+is gone.
```
Gozala, line 59: I'm not sure I understand the value of creating account keypair and burning in as opposed to using current model where it passkey derived. We still could delegate to a recovery key. And delegate to other passkey.

Main difference is that if passkey somehow gets compromised it could revoke delegation to recovery and other passkeys but given non extractable nature of passkeys risk seems negligible. Also delegation to second passkey would still happen through other passkey (or recovery key) so I'm not sure we gain much other than forming tree shape over dag shape.

docs/passkey-fanout.md
```diff
@@ -0,0 +1,274 @@
...
+operations:
+
+- **publish** a signed delegation chain;
+- **claim** every chain addressed to a given key.
+
+Both are unauthenticated, and this is not a concession. The entries are
+self-authenticating UCAN delegations: publishing one confers nothing, and
+knowing a DID is not authority, since a delegation is useless without the
+matching private key. A store can withhold an entry. It cannot forge one,
+alter one, or grant anything by holding one.
```
Gozala, line 109: I think allowing anyone to list delegations to specific DID is a risk, why not require signing that request instead ?

Gozala (commented):
Thanks for writing this up. I think it makes sense although I'm not necessarily convinced that having special account key improves security enough to justify added complexity. It also prevents us to defer from setting up recovery at later point in time.

I think real solution to accounts is to represent them as `did:web` identifiers that users can either bring with them or create them with tonk when openning an account. This will enable key rotation and off band recovery without invalidating delegation chains etc...

E.g we could allow users to obtain `did:web:network.tonk.gozala` like identifier once they prove email ownership with current flow. When user want to recover access to the account because they lost underlaying keys they could proved email ownership and hand us new key. This seems more honest solution to the whole problem and identifier will reflect actual custody that is `tonk.network`.

We could also in the future add support for did:plc or did:mailto and maybe other identifiers for accounts in which case recovery becomes key rotation and their problem.

I also think that we can start with passkey generated did:key identifiers

PR #714, opened 2026-08-18
Consumer provisioning via /provider/add

Gozala:
Stacked on #713. Second increment of the [access metering plan](https://github.com/tonk-labs/tonk/pull/702): space provisioning, so the service knows which customer pays for which space before metering attributes anything.

`/provider/add` is a capability in `tonk-account::customer` alongside enroll/activate (`Provider` attenuation, so delegating `/provider` can never leak `/customer` or `/sponsor` powers), invoked on the customer's own subject with the space's consent deposited by CID. The consent is the powerline delegation the space already mints to the account at creation, exactly as the plan predicted: the server validates it as the consumer's own delegation to the invoking customer granting `/consumer/provision` or broader (audience is what binds it), and provisions idempotently, `ON CONFLICT` keeps re-adding under the same provider a no-op while a consumer someone else provides answers `ConsumerProvided`. Activation is not required to add, only to serve: consumers added while the provider is `Registered` go live with its activation.

Client side, provisioning happens where retention already does: the worker at space creation next to `retain_space_delegation`, the CLI at spot bootstrap, both best effort for the same reason retain is, a space is usable the moment its delegations exist locally. Both build the container through the shared `tonk_identity::request::build_provider_add_invocation`.

Service tests cover the happy path (consent = powerline, Registered provider), idempotent re-add, second-provider refusal, and a consent issued to a different customer.

---

## PR-Codex overview
This PR focuses on enhancing the consumer provisioning process in the `tonk` system, introducing new functionalities for handling consumer spaces under providers, along with related error handling and database operations.

### Detailed summary
- Changed return type of a function in `registration.rs` from `Receipt` to `Answer`.
- Added `add_consumer` function in both `sqlite.rs` and `d1.rs` for provisioning consumers.
- Introduced a new command path segment `/provider/add` and its handling in `registration.rs`.
- Enhanced error handling for consumer provisioning scenarios.
- Added tests for consumer provisioning, including cases for successful and refused provisioning.
- Updated `Answer` enum to differentiate between customer and consumer receipts.
- Implemented verification for consent in the provisioning process.

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

PR #713, opened 2026-08-18
Customer enrollment and activation in the access service

Gozala:
Implements customer registration from the [access metering plan](https://github.com/tonk-labs/tonk/pull/702) end to end: the service side (enroll/activate at `POST /ucan/`), and the browser side that replaces the email-code signup ceremony. Space provisioning (`/provider/add` + `/consumer/provision`) follows as the next increment.

## Service side

Both commands land at `POST /ucan/` as ordinary invocation containers, dispatched by a command peek ahead of the presign path, which stays byte-identical for everything else. The capability types (`Customer` attenuation, `Enroll`/`Activate` effects) live in `tonk-account::customer` so clients build invocations from the same definitions the service decodes with; the service executes them through `dialog_capability::Provider` environments over a `Store` trait with D1 and sqlite twins. `Store` and `EmailSender` are declared through dialog's dual `async_trait` forms, so one generic `Provider` impl serves every environment.

Enrollment verifies the chain rooted at the customer's self-certifying DID, validates the deposited access delegation (an argument, not a proof, and acceptable as a chain: the device issues it under its `root → device` grant), writes the customer and its self-provided account consumer in one batch, and emails an activation link. The link embeds a **complete service-signed invocation**: presenting it is activating, so no key is needed on the clicking device and cross-device activation needs no relay. Activation enforces the invocation's `exp` against the clock (the pinned dialog revision's `verify` checks structural coherence, not the present), records the terms version baked in at mint time, and lands the customer on the trial plan.

The service gains a signing identity from a `SERVICE_SECRET_KEY` secret, `did:key` today, published at `/.well-known/did.json` under the host's `did:web` name and as `serviceDid` in `/.well-known/tonk` for discovery. `GET /customer/:did` answers registration state for client polls. The Resend transport moves to a shared `tonk-email` crate.

## Browser side

Signup becomes email → passkey → attach → enroll: the code request/verify panels are gone, and the account service accepts uncoded creation (legacy clients that still send a code get it validated). After attach the worker builds the enroll invocation with the device key, deposits a device-issued access delegation chained through the `root → device` grant, and POSTs same-origin `/ucan/`; login enrolls the same way using the account's recorded email, with an already-active customer answered quietly. The dashboard surfaces a pending activation from `GET /api/customer`, and the emailed link opens `/activate`, a new top-document page whose accept button presents the embedded invocation.

[...]

PR #715, opened 2026-08-18
Meter invocations into an ingest database

Gozala:
Stacked on #714. Phase 0 (meter, no enforcement, no billing) of the [access metering plan](https://github.com/tonk-labs/tonk/pull/702): the full write path for activity, so calibration data accumulates while everything else stays as it was.

`POST /ucan/` now records one row per attributable invocation into a second D1 database (`INGEST`), deliberately separate from control: bulky, write-heavy, disposable once charged, with its own 10 GB ceiling. Permits record with the declared write bytes read off the signed `Content-Length`; 401/403 refusals record as `outcome = denied` with the refusal's kind, since a client retrying against a blocked consumer still costs invocations; infra failures and unparseable containers are the service's cost and are not attributed. The invocation bytes ride inline, and the presented proof set is flattened once per operator session (`INSERT OR IGNORE` on `chain`/`block`), keyed by a hash over the sorted proof CIDs so token order does not split sessions.

The endpoint moves out of the Router onto the fetch event: recording must outlive the response, and only the event's `Context` provides `wait_until`. The authorizer and registration handlers take `Env` directly as a result. A metering failure logs and never costs the caller the permit they already hold, and a missing `INGEST` binding (pre-bootstrap deploys) skips metering rather than refusing service.

The native helpers server mirrors the same collection into sqlite with a `/_test/ingest` probe, and the push/pull integration test asserts its permits landed as rows. CI provisions `tonk-access-ingest-pr-<N>` per preview alongside the control database; staging/production need a one-time `wrangler d1 create tonk-access-ingest{-staging,}` with ids filled into `wrangler.toml`, added to the bootstrap list from #713.

Next in the stack: the charge cron (usage aggregation → ledger) per plan §10.

---

## PR-Codex overview
This PR introduces an `ingest` module to handle invocation records in a separate database, enhancing the metering functionality. It includes new database schemas, updates to various handlers, and adjustments to ensure proper recording of invocation metrics.

### Detailed summary
- Added `ingest` module for handling invocation records.
- Introduced new SQL tables for `invocation`, `chain`, and `block`.
- Updated `main` function to handle POST requests for `/ucan/` and record invocations.
- Enhanced error handling in the `handle` function.
- Modified `record_invocation` to log invocation metrics.
- Updated `registration` logic to support multiple access deposits.
- Implemented scoped deposits for customer enrollments.
- Adjusted tests to verify new functionality and ensure correctness.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -149,38 +149,51 @@ jobs:
       # The access worker's control database follows the same per-PR shape
       # as the account worker's, and for the same reason: divergent
       # migration histories cannot share a database.
-      - name: Prepare preview access database
+      - name: Prepare preview access databases
         if: github.event_name == 'pull_request'
         id: access-database
         run: |
           set -euo pipefail
-          database="tonk-access-control-pr-${{ github.event.pull_request.number }}"
-          config="$PWD/wrangler.preview.toml"
-          database_id="$({
-            nix --accept-flake-config develop .#ci --command wrangler d1 list --json
-          } | jq -r --arg name "$database" '.[] | select(.name == $name) | .uuid' | head -n 1)"
-          if [ -z "$database_id" ]; then
-            nix --accept-flake-config develop .#ci --command \
-              wrangler d1 create "$database" --location weur
-            database_id="$({
+          resolve() {
+            local database="$1"
+            local id
+            id="$({
               nix --accept-flake-config develop .#ci --command wrangler d1 list --json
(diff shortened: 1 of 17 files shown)


Gozala:
Superseded by #724, which lands this stack in one merge — branch protection's strict up-to-date rule plus the stacked-PR merge restrictions made per-PR landing impractical. All content and history preserved there; this PR remains the review record for its slice.

PR #722, opened 2026-08-18
fix(cli): restore legacy commands and rules during migration

jackddouglas:
## Summary

- retain legacy `dialog.concept/transient` markers and `dialog.effect/{source,polarity}` as migration sidecar metadata
- rewrite embedded `dialog.origin/*` rule attributes, compile effects into native `dialog.rule/*` facts, and restore current boolean command markers after import
- validate real effects before mutating the destination, ignore only non-`effect:` test debris, and report restored command/rule counts
- add a regression proving a migrated transient command still fires its rule

## Root cause

The CSV rewrite correctly excluded the reserved `dialog.*` rows, but assumed the current runtime would regenerate them. It cannot regenerate user-authored command markers or effects. The migrated Vault concepts and data therefore rendered, while `vault/open` resolved as durable and had no installed handler rule, so clicking a sidebar node never produced `vault/here`.

## Verification

- `cargo test -p tonk-cli legacy::tests --lib --no-fail-fast` (8 passed)
- `cargo test -p tonk-cli --features integration-tests --test legacy_migration --no-fail-fast -- --nocapture` (2 passed, 1 network-gated test ignored)
- `cargo clippy -p tonk-cli --all-targets --all-features -- -D warnings`
- live tonk-team migration: 404/404 blobs verified; 25 command markers and 53 native rules restored; one invalid non-effect test entity ignored
- live `vault/open` dispatch created `vault/here.active = id:vault/n/claude-md-2d5d10`; repaired staging head synced at `#4eUtRnc3xsLkjJbXTi8fiUES7KZyB3Qehnin8iqkFwkr`

## Separate issue

The legacy account repository still reports `missing field branch`; that account hydration failure is independent of the repaired tonk-team repository and Vault command path.

---

## PR-Codex overview
This PR focuses on enhancing the migration process of legacy commands and effects in the `tonk-cli` by introducing a new method to import upgraded branches, translating legacy runtime metadata, and ensuring that transient commands and rules are correctly restored.

### Detailed summary
- Introduces `import_upgraded_branch` function to handle legacy command and effect migration.
- Adjusts error handling for importing branches.
- Updates logging messages for clarity during migration.
- Adds tests for migrated transient commands and effects.
- Modifies `Upgraded` struct to include `legacy_csv`.
- Implements runtime repair logic for legacy commands.
- Updates legacy runtime extraction and repair methods.
- Enhances error handling for malformed legacy effects.
- Ensures unmarked legacy effects are ignored during migration.

### DIFF
--- a/rust/tonk-cli/src/bin/tonk.rs
+++ b/rust/tonk-cli/src/bin/tonk.rs
@@ -2000,19 +2000,29 @@ async fn legacy_migrate(
             Ok(blobs) => blobs,
             Err(error) => return print_failure(error),
         };
-        if let Err(error) =
-            tonk_cli::legacy::import_migrated_branch(&destination, branch, &upgraded.csv).await
+        let repaired = match tonk_cli::legacy::import_upgraded_branch(
+            &destination,
+            branch,
+            &upgraded.csv,
+            &upgraded.legacy_csv,
+        )
+        .await
         {
-            return print_failure(error);
-        }
+            Ok(repaired) => repaired,
+            Err(error) => return print_failure(error),
+        };
         println!(
             "{branch}: {} rows ({} remapped, {} dropped as dialog's own), \
-             {} blobs ({} bytes)",
+             {} blobs ({} bytes), {} commands and {} rules restored \
+             ({} invalid legacy effect entities ignored)",
             upgraded.migration.kept,
             upgraded.migration.remapped,
             upgraded.migration.dropped,
             blobs.copied,
-            blobs.bytes
+            blobs.bytes,
+            repaired.transient_concepts,
+            repaired.native_rules,
+            repaired.ignored_effects
         );
     }
     println!("upgraded {} branch(es)", branches.len());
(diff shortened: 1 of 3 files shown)


Gozala:
@jackddouglas I'm not sure I fully understand what is going in is it that `tonk schema` does not properly capture command! definitions or does new dialog does not evaluate command! definitions properly ?

Gozala:
> @jackddouglas I'm not sure I fully understand what is going in is it that `tonk schema` does not properly capture command! definitions or does new dialog does not evaluate command! definitions properly ?

I was wrong, I thought I did full schema to yaml export and then import after data was migrated, but that didn't got implemented which is why the issue described here manifests.

I'm actually not sure this will address everything as the way dialog now stores and indexes concepts and rules have changed beyond just attribute namespace and I suspect it may still not fully fix all the issues. Great if it does.

I am exploring implementing exporting definitions into yaml so it could be then evaluated as part of migration to make the whole thing less brittle

jackddouglas:
i think you're right about the conceptual boundary and we should do it for future migrations but that abstraction doesn't exist in cli 0.6.7 which is why this pr exists. since the blast radius is small and this is a very specific migration i'd rather finish out this pr to have something that works for the few spaces we care about and implement the definition yaml export for the future.

PR #733, opened 2026-08-20
chore: move staging onto tonk.network

Gozala:
#730 moved production to the `tonk.network` apex but left staging on `staging.tonk.xyz`, so the two deployments now straddle different registrable domains. That split is the kind of thing that quietly rots: anything reasoning about "our origin" has to know about two of them, and the analytics environment tag had already drifted as a result. This moves staging onto the same domain as production.

Routes become `staging.tonk.network` and `accounts-staging.tonk.network`. The access service's staging vars follow, since they name the account service directly. Custom Domains provision their own DNS record and certificate on deploy, so there is no DNS step to do by hand.

The passkey RP ID is the reason to be careful here, and it survives the move intact. `apex_rp_id` matches the apex exactly (`host == RP_APEX`), so `staging.tonk.network` is off-apex and mints its own credentials rather than production's — the same property `staging.tonk.xyz` had, now holding for a subdomain of the apex rather than a separate domain. The test that pinned this was asserting it about the retired hostname, so it now asserts it about the one that actually ships.

One change is not cosmetic: `tonk-analytics` decides the PostHog `environment` super-property by comparing `location.hostname` literally. Left alone, every staging event would have been tagged `dev` and mixed into the dev bucket in the shared project.

## Deployment note

The WAF rate limiting rule for the account service does not carry over. It is scoped to zone `tonk.xyz` and matches on `http.host`, so moving staging to `tonk.network` puts it in a different zone and the rule stops applying. It is the only per-IP limit on unauthenticated `POST /codes`, `/accounts/preflight`, and `/links` — the service enforces a per-email cooldown but nothing per-IP, and staging shares production's Resend key. It needs recreating on the `tonk.network` zone at deploy time; the account service README now says so.

## Verification

- `cargo fmt --all` clean
- `cargo clippy --all --all-targets --all-features -- -D warnings` clean
- `cargo test -p tonk-identity -p tonk-analytics` passing
- `cargo test --no-run --target wasm32-unknown-unknown -p tonk-identity -p tonk-analytics` compiles, so the wasm-gated RP-ID test is not stale
- Both wrangler configs parse and resolve to the intended routes
- Two `tonk-cli` `when_migrating_from_carry` failures are pre-existing on `staging` (malformed stored account-session state, `missing field pending_detaches`) and unrelated; confirmed by running them against an unmodified tree
- No CI changes needed: `publish.yml` deploys via `--env staging -c wrangler.toml` and reads the route from config, and `.github/` contains no hostname literals

---

## PR-Codex overview
This PR focuses on updating references from `staging.tonk.xyz` to `staging.tonk.network` across various files to ensure consistency in the staging environment URLs.

### Detailed summary
- Updated documentation comments in `rust/tonk-host/src/space_origin.rs`.
- Changed assertions in `rust/tonk-identity/src/passkey.rs`.
- Modified route patterns in `wrangler.account.toml` and `wrangler.toml`.
- Updated environment descriptions in `docs/telemetry.md`.
- Adjusted comments in `rust/tonk-analytics/src/web.rs`.
- Changed URLs in `rust/tonk-cli/src/invite.rs`.
- Updated service URLs in `rust/tonk-access-service/README.md` and `rust/tonk-account-service/README.md`.

### DIFF
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -36,7 +36,7 @@ before a profile exists report as `tonk:anonymous`.
 
 Every web event additionally carries an `environment` super property
 derived from the hostname: `production` (tonk.network), `staging`
-(staging.tonk.xyz), or `dev` (anything else). The marketing site's
+(staging.tonk.network), or `dev` (anything else). The marketing site's
 own snippet registers `environment = "website"` in its repo, so all
 surfaces share one filterable dimension within the PostHog project.
 
--- a/rust/tonk-access-service/README.md
+++ b/rust/tonk-access-service/README.md
@@ -78,7 +78,7 @@ The S3 address uses region `auto`, as R2 requires.
 
 As a Cloudflare Worker, build and deploy with `worker-build` / `wrangler` like any `worker`-based crate (the `cdylib` target).
 
-`wrangler.toml` at the repo root carries three environments. The top level is production on `tonk.network`, `[env.staging]` is `staging.tonk.xyz`, and `[env.preview]` is on no route: the deploy workflow uploads one version of it per pull request under a `pr-<number>` preview alias, reached at a workers.dev URL. Each has its own R2 buckets, because this Worker presigns writes into whichever bucket it is bound to and a preview must not be able to write into a real one.
+`wrangler.toml` at the repo root carries three environments. The top level is production on `tonk.network`, `[env.staging]` is `staging.tonk.network`, and `[env.preview]` is on no route: the deploy workflow uploads one version of it per pull request under a `pr-<number>` preview alias, reached at a workers.dev URL. Each has its own R2 buckets, because this Worker presigns writes into whichever bucket it is bound to and a preview must not be able to write into a real one.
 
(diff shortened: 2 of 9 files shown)


Gozala:
Dry-run of the staging environment completed (exit 0). It validated 979 assets and confirmed the staging bindings resolve to staging resources only:

```
env.CONTROL (tonk-access-control-staging)                D1 Database
env.INGEST (tonk-access-ingest-staging)                  D1 Database
env.BUCKET (tonk-spaces-staging)                         R2 Bucket
env.REVOCATIONS (tonk-revocations-staging)               R2 Bucket
env.ACCOUNT_SERVICE_URL (https://accounts-staging.tonk.network)
env.REVOCATION_RELAY_URL (https://accounts-staging.tonk.network/revocations)
```

jackddouglas:
> The passkey RP ID is the reason to be careful here, and it survives the move intact. apex_rp_id matches the apex exactly (host == RP_APEX), so staging.tonk.network is off-apex and mints its own credentials rather than production's — the same property staging.tonk.xyz had, now holding for a subdomain of the apex rather than a separate domain. The test that pinned this was asserting it about the retired hostname, so it now asserts it about the one that actually ships.

the reason i had staging off `tonk.xyz` is so we could have it on an entirely different apex, that way if we wanted to expand the scope of the stable passkey rp id we could do so without affecting staging's isolation. could we expand the scope in the future to wildcard subdomains for example without affecting staging here?

PR #734, opened 2026-08-20
test(cli): avoid revision-hash collision in live migration assertion

jackddouglas:
## Summary

- assert live-migration query results through structured matches instead of rendered stdout
- verify the derived pong row and its tag directly
- verify the transient ping query returns zero rows

## Root cause

The staging failure was a false positive. The test searched the entire rendered response for `hi`; the failing response had zero claims, but its revision hash contained `hi` (`#5aivuLX7hi7FF...`).

## Verification

- `cargo fmt --all -- --check`
- `nix develop path:. -c cargo test -p tonk-cli --test live_migration -- --nocapture` (3 passed)
- `nix develop path:. -c test:native:debug` (1,497 passed, 4 skipped)

---

## PR-Codex overview
This PR focuses on enhancing the assertions in the test for live migration in the `tonk-cli` by ensuring that the derived `pong` retains the correct tag and that the `ping` command does not persist after execution.

### Detailed summary
- Changed the assertion for `pong` to check the length of results and ensure exactly one result is derived.
- Updated the assertion for `pong` to confirm the derived tag matches the expected value.
- Modified the assertion for `ping` to check that no results persist after execution.

### DIFF
--- a/rust/tonk-cli/tests/live_migration.rs
+++ b/rust/tonk-cli/tests/live_migration.rs
@@ -74,21 +74,29 @@ concept!: &pong
     // The derived fact is durable...
     let pong = test
         .eval_inline("pong:\n  this: ?this\n  tag: ?tag\n")
-        .await?
-        .stdout;
-    assert!(
-        pong.contains("hi"),
-        "the rule must derive a durable pong from the ping command; saw:\n{pong}"
+        .await?;
+    let pong_results = &pong.response.matches_after[0].results;
+    assert_eq!(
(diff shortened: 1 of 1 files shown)


jackddouglas:
Closing as superseded by #754. Staging now checks the rendered field `tag: "hi"` rather than a bare `hi`, so a revision hash containing those two letters cannot produce the false positive this PR fixed. The exact regression is covered without needing this alternate structured assertion.

PR #748, opened 2026-08-23
Spec-compliant UCAN revocations, verified through dialog

Gozala:
Revocation was hand-rolled here: a bespoke artifact shape, a `target_cid`/`path_hex` pair invented for it, and a screen that ran after authorization and re-derived the chain's structure from a second parse. Each piece was reasonable alone, and together they added up to a second implementation of something the UCAN spec and dialog both already describe.

So the whole thing moves onto dialog's model. Minting and verification go through `RevocationChain`, which checks the invocation's own validity, the witness path's linkage and rooting, and that the revoker held what it revokes — every one of those a check this crate previously hand-rolled and got subtly wrong somewhere. The service now only answers two questions of its own: do we hold data this revocation could protect, and has it been recorded already.

The enforcement side moves inside the chain walk rather than sitting after it. That is not just relocation: a screen outside the walk sees one flat set of issuers and has to apply it to every hop, which for `a -> b -> c -> d` lets d's issuer revoke c — a principal that merely *received* authority withdrawing the grant it depends on. Dialog scopes the candidates per link (the issuers at or above it, plus that link's own audience, who may always disclaim what it was given), so wiring the KV index in through `with_revocations` tightens the check as well as moving it. A test pins exactly that difference, and fails if the candidate set goes flat again.

Two things this exposed and fixed on the way:

- **`sub` names the revoker, not the revoked capability.** Those became different DIDs once the artifact followed the spec, and the consumer lookup was silently following the wrong one — asking whether a *device* was a paying consumer when the question is whether we hold data for the *space*. Device grants are `Subject::Any` powerlines, so the revoked subject falls back to the issuer.
- **Refusals arrived unreadable.** The native server rendered every authorization failure into `Authorization failed: {e}`, and dialog itself flattened every typed reason into `Malformed` before the boundary saw it. A caller could not tell an expired proof from a revoked one — the distinction its retry logic turns on. Fixed at the source in dialog (dialog-db#461) rather than worked around, which then let ~100 lines of window-screening delete outright: the walk both enforces the window and now names the refusal.

The e2e asserts the property that matters and the one the device list cannot show: after revocation the guest's writes no longer reach the owner. Asserted on content rather than status codes, with a positive control proving the write path worked beforehand — earlier versions of this test were vacuous, since a replica with nothing to send answers 200 either way.

Depends on dialog-db#460 and dialog-db#461; the dialog pin returns to a dated tag before merge.

---

## PR-Codex overview
This PR focuses on restructuring the revocation handling in the `tonk` project, including the removal of outdated revocation-related files and the introduction of a new `revoke` module. It also enhances the revocation publishing mechanism and adjusts related configurations.

### Detailed summary
- Deleted multiple files related to revocations and expiry.
- Added a new `pub mod revoke;` in `rust/tonk-access-service/src/handlers.rs`.
- Introduced a new test for revocation in `rust/tonk-cli/Cargo.toml`.
- Updated `rust/tonk-access-service/src/handlers/config.rs` to remove revocation URL.
- Removed revocation URL references in various modules.
- Enhanced revocation handling in `rust/tonk-worker/src/router.rs`.
- Implemented a new `publish_revocation` function to ensure revocations are published across services.
- Adjusted tests to reflect changes in revocation handling.
- Updated documentation and comments to clarify the new revocation structure.

> The following files were skipped due to too many changes: `rust/tonk-account-service/src/helpers/server.rs`, `rust/tonk-access-service/src/revocation/checker.rs`, `rust/tonk-access-service/src/helpers/server.rs`, `rust/tonk-access-service/src/handlers/ucan.rs`, `rust/tonk-fab/src/share.rs`, `rust/tonk-cli/tests/invite_revocation.rs`, `rust/tonk-access-service/README.md`, `rust/tonk-fab/src/logic.rs`, `rust/tonk-account-service/src/core/devices.rs`, `Cargo.lock`, `rust/tonk-ui/src/account_flow.rs`, `rust/tonk-worker/src/router/create_invite.rs`, `rust/tonk-access-service/src/revocation/index.rs`, `rust/tonk-access-service/src/revoke.rs`, `rust/tonk-access-service/tests/ucan_integration.rs`, `rust/tonk-worker/src/router/repository.rs`, `rust/tonk-worker/src/router/revoke_invite.rs`, `rust/tonk-access-service/src/revocation.rs`, `plan/revocations-at-ucan.md`, `rust/tonk-identity/src/revocation.rs`


Gozala:
Correcting something in this branch's history, since it's visible in the commits: `728ccff5` justified keeping the window screen by saying clients depend on the `Expired` code. I checked afterwards — nothing in tonk or dialog reads that variant today, so that reasoning was wrong.

The screen still had to go, and for a better reason than the one I gave: `bf9a55b9` removed it because the chain walk now both enforces the window and names the refusal identically. That's pinned by `it_answers_a_refusal_with_its_typed_reason`, which asserts the refusal through `AuthorizeError` deserialization — the same path `read_rejection` takes — rather than on the JSON shape, so it fails if the body stops being readable by a real client.

PR #759, opened 2026-08-25
Custody space and invite seeds under the account's X25519 key

Gozala:
The plans (`onboarding-accreditation.md`, `authority-facts.md` step 3) call for every space and invite seed to be custodied under the account so a space can be re-issued after rotation or a lost device. Nothing did that: a created space's secret lived only in the creating device's credential store, and an invite principal was consumed for one hop and dropped.

Wrapping under the account KEK would have made space creation a browser-only act, since the KEK only exists inside a passkey ceremony and the CLI never runs one. So seeds are sealed to a **public** X25519 key instead, derived from the account secret under the already-reserved `tonk/enc/v1` context and published as a `did:key` under the `x25519-pub` (`0xec`) codec. Any device holding the account DB can seal; only a ceremony can open.

- `tonk-identity::sealed`: `EncryptionKey` / `RecipientKey` and a `Sealed` envelope (ephemeral X25519 + HKDF + AES-256-GCM, recipient and subject DIDs bound as associated data).
- `tonk-schema`: `AccountEncryptionKey` on the account subject; `CustodiedSeed`, one row per `(subject, recipient)` with subject, kind, and recipient repeated as queryable attributes so rotation and recovery can enumerate rows.
- Creation and unlock ceremonies return the recipient with the root; the worker keeps it on the local root record (a re-save from a ceremony that did not hold the secret inherits it) and publishes it on the account sweep.
- Space create seals its seed (`tonk:space`); a durable join through an open invite seals the invite principal's seed (`tonk:invite`), since the membership hangs off that principal and the joiner's account is what re-issues it at rotation. Both are best-effort like `retain_space_delegation`.
- The space signer now imports from a seed drawn up front, so the stored credential keeps its shape; an extractable WebCrypto key stored as the credential cannot be reloaded.

Not in this PR: the CLI writer (CLI spaces do not delegate to the account yet), the backfill sweep for spaces created before the key existed, and rotation/recovery consuming the rows (authority-facts step 4). Plans updated to match.

---

## PR-Codex overview
This PR introduces support for X25519 encryption keys in the `tonk` project, enhancing security by allowing devices to seal seeds to a public key. It includes updates across various modules to incorporate the new encryption functionality and related data structures.

### Detailed summary
- Added `x25519-dalek` dependency in `Cargo.toml`.
- Introduced `sealed` module in `tonk-identity`.
- Updated `AccountEncryptionKey` structure in `tonk-schema`.
- Incorporated `encryption_key` in multiple structs and functions across modules.
- Implemented sealing and opening of seeds using X25519 keys.
- Enhanced tests to ensure encryption key functionality and validation.
- Updated relevant documentation to reflect changes in key management.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5036,6 +5036,7 @@ dependencies = [
  "aes-gcm",
  "anyhow",
  "blake3",
+ "bs58",
  "dialog-common",
  "dialog-credentials",
  "dialog-ucan-core",
@@ -5058,6 +5059,7 @@ dependencies = [
  "wasm-bindgen-futures",
  "wasm-bindgen-test",
  "web-sys",
+ "x25519-dalek",
  "zeroize",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -181,6 +181,7 @@ web-time = "1"
 webbrowser = "1.0"
 worker = "0.8.5"
 worker-macros = "0.8.5"
+x25519-dalek = { version = "2", features = ["static_secrets"] }
 zeroize = "1"
 
 js-sys = "0.3"
--- a/plan/authority-facts.md
+++ b/plan/authority-facts.md
@@ -131,33 +131,109 @@ devices.
 The worker's label is coarser — no `platform`, no touch-point count — so
 browser and OS come from the user agent alone.
 
-## Wrapped keys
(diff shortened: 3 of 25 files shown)


Gozala:

PR #763, opened 2026-08-25
Every join is durable to the device's account; the guest path goes

Gozala:
Stacked on #759. Stage 1 of `plan/join-under-custody.md`.

A guest visit was an accountless join: a one-hour hop from the invite principal to the device operator, the invite URL retained on disk as the bearer to replay, a renewal path in sync, a promote button in the bar, and a `PendingIntent` round trip through `/account` for the durable join that had to follow. There are no accountless devices any more (#745), so all of that was a second join with weaker authority kept alive by a stale assumption, and it sent registered users through it too.

Now every join is durable, to whatever account the device has: `current_account()` yields the passkey root and its device grant when one is linked, or the onboarding account and its powerline otherwise, minted on first use. The claim terminates there and the invite principal's seed is sealed to that account (`CustodiedSeed`, from #759), so accreditation can re-root the membership rather than the user redoing the join. The invite URL is never kept.

Removed: `JoinMode`/`JoinRejection`, the guest record/lease/grant and promotion, `/api/profile/visit`, `/api/repository/{repo}/membership`, `notify_account_required` and `PendingIntent` (`DurableJoin` and `CreateSpace` alike, since creating a space under the onboarding account needs no gate either), `require_account`, the `AccountRequired` error variants, the FAB's join button, join-first dialog and promote request, the UI's parked-intent gate (only `next` handling stays), and the guest-only tests. The FAB's membership probe becomes a presence probe on `GET /api/repository/{did}`, keeping the "space not held here" state.

Verified in Chrome: tonk-worker 288/288, tonk-fab 47/47 plus integration binaries, tonk-ui account_gate; native clippy `-D warnings` clean on all four crates. Route table on #760 loses two entries when this lands.

Stage 2 (public-key spaces) and Stage 3 (accreditation as rotation) follow on top.

---

## PR-Codex overview
This PR refactors the handling of user accounts and guest visits in the `tonk` project. It eliminates the concept of guest visits, ensuring all joins are durable and associated with an account from the start.

### Detailed summary
- Removed `tonk_ui::account_gate::install()`.
- Deleted guest-related constants and types, including `BLOCKED_NEEDS_MEMBERSHIP`, `VisitRequest`, `VisitResponse`, and `MembershipResponse`.
- Changed API endpoints from `/api/profile/visit` to `/api/profile/join`.
- Updated functions to reflect the removal of guest visit functionality.
- Modified account-related navigation and session handling to remove guest logic.
- Adjusted tests and UI components to align with the new durable-only join process.

> The following files were skipped due to too many changes: `rust/tonk-fab/src/element.rs`, `rust/tonk-worker/src/router/join.rs`


Gozala:

PR #764, opened 2026-08-25
Spaces are created with a public key; the seed lives under custody

Gozala:
Stacked on #763. Stage 2 of `plan/join-under-custody.md`.

A space's private key used to be stored in the creating device's credential store, a copy the account could never revoke and the reason accreditation needed that device. Now the signer exists only long enough to mint `space → account`; the repository stores the verifier, and every later act on the space proves through `space → account → device`, the way a joined replica already does. The only copy of the space secret is the `CustodiedSeed` sealed to the account (#759), written before anything durable; a seed that cannot be custodied is a space that is not created.

`adopt_profile_spaces` re-issues an onboarding-rooted space to the passkey root by opening its custodied seed with the onboarding secret rather than reaching for a local signer. Re-sealing to the new account, retracting the old row, and retiring the onboarding account are Stage 3.

Two consequences worth knowing:
- The encryption-key read no longer needs a ready account branch: the fact lives on profile `main` and is read as-is.
- Creating a space on a linked device requires the account's published encryption key. New accounts publish it at creation; existing accounts publish it the first time they unlock on any browser (#759). Until then a linked device refuses to create a space with a clear message rather than minting a key it cannot custody.

Verified in Chrome: tonk-worker 288/288 plus the two new tests (verifier-only credential, seed opens to the space's DID and the space still proves; onboarding space re-issued to the root from its seed). Native clippy `-D warnings` clean.

---

## PR-Codex overview
This PR focuses on enhancing the handling of account encryption keys and the management of spaces within the Tonk system, particularly regarding the persistence of keys and the adoption of onboarding spaces.

### Detailed summary
- Added `persist_test_root` function to persist the test root state.
- Updated `test_state_without_account` to call `persist_test_root`.
- Introduced `published_encryption_key` to read the encryption key for accounts.
- Modified `custody_recipient` to publish encryption keys when necessary.
- Created `open_onboarding_seed` to access seeds custodied for onboarding accounts.
- Updated `create_repository` to ensure space seeds are custodied under accounts.
- Enhanced `adopt_profile_spaces` to re-issue spaces from custodied seeds.
- Added tests for space creation and onboarding space adoption scenarios.

### DIFF
--- a/rust/tonk-worker/src/router.rs
+++ b/rust/tonk-worker/src/router.rs
@@ -592,24 +592,57 @@ pub mod tests {
     /// creates a root without an account around it.
     pub async fn test_state_without_account() -> TonkState {
         let state = test_state_without_root().await;
+        persist_test_root(&state).await;
+        state
+    }
+
+    /// Persist the test root on `state`, the way a creation or unlock
+    /// ceremony does: the `root -> device` grant, the recipient custodied
+    /// seeds are sealed to, and that recipient published on profile main.
+    /// Returns the root DID.
+    pub(crate) async fn persist_test_root(state: &TonkState) -> dialog_varsig::Did {
         let root = Ed25519Signer::import(&test_root_seed(&state.profile_name))
             .await
             .unwrap();
+        let root_did = root.did();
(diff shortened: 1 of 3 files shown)


Gozala:

PR #765, opened 2026-08-25
Accreditation rotates the onboarding account's custody to the passkey account

Gozala:
Stacked on #764. Stage 3 of `plan/join-under-custody.md`; the passkey is needed for nothing here.

When a passkey account lands on a device, `rotate_from_onboarding` runs where `adopt_profile_spaces` used to. It opens every `CustodiedSeed` sealed to the onboarding account with the onboarding secret (local, no ceremony) and mints a fresh `subject → root` from the seed: a created space from its own signer, exactly as creation does; a joined membership by rebuilding the stored chain with the `principal → onboarding` hop replaced by `principal → root`, so nothing is appended below the onboarding account and no chain keeps a hop at it. Each seed is re-sealed to the account's published key and its old row retracted. Once nothing is left under the onboarding account, its grant to the device is revoked (minted by the onboarding account itself, published best-effort) and its custodian is demoted to the public half, leaving the envelope unopenable; `onboarding::account` then errors rather than minting a second onboarding account.

Resumable: seeds are independent, one that fails stays sealed to the onboarding recipient and is picked up next time, and retirement waits for it. Deferred, with a log line, until the account has published its encryption key (#759), which the sweep does before this runs.

Verified in Chrome: tonk-worker 290/290, including a new test that creates one space and joins another under the onboarding account, accredits, and checks both prefixes end at the root with no onboarding hop, both still prove, both seeds are sealed to the new recipient and none to the old, and the onboarding account can no longer be opened. Native clippy `-D warnings` clean.

What this does not do yet: rotate a passkey account to a new secret (same function, with a ceremony-supplied key; authority-facts step 4), and remove the onboarding `DeviceLink` row from the device list.

---

## PR-Codex overview
This PR focuses on enhancing the onboarding account management by introducing functions to rotate custody from the onboarding account to the passkey account and retire the onboarding account, ensuring that all associated seeds and delegations are properly managed.

### Detailed summary
- Changed `retain_claim_authority` and `publish_revocation` functions to `pub(super)`.
- Added a new function `retire` to manage onboarding account retirement.
- Introduced `rotate_from_onboarding` to transfer custody from onboarding to passkey account.
- Updated comments and documentation for clarity.
- Removed deprecated function `open_onboarding_seed`.
- Adjusted tests to validate new custody rotation functionality.


Gozala:

PR #772, opened 2026-08-26
The custody publish is pre-signed at the ceremony and drained by the worker

Gozala:
Live testing kept finding the same shape of problem: passkey prompts with no context, and a "finish backup" interaction that existed only because the custody-cell publish was blocked until email activation while its signature needed the passkey.

This PR removes the interaction instead of decorating it. The creation ceremony — the one moment the PRF-derived custody key is in hand — now pre-signs the `/memory/publish` invocation (bounded to thirty days) and hands it back with the sealed cell. Both ride the pending queue, and the worker drains them itself on activation, on boot, and when new work queues: no page, no assertion, no button. The old page-side republish, its bridge, the backup button, and the completion route are deleted. The dashboard keeps only a passive "Finishing your account's backup…" note while the queue is non-empty.

The sealed cell is also recorded on profile main as a `CustodyCell` fact: the account's own sync then carries the recovery envelope to every device holding the profile, and the vault copy remains what bootstraps a brand-new browser.

Assertions that do still happen are pinned: the worker's encryption-key request and the post-create PRF chases set `allowCredentials` from the recorded credential, so a browser holding several passkeys for this origin cannot answer with one that derives a different custody space. Login and unlock stay discoverable — choosing the passkey is choosing the account. The worker's webauthn request raises a consent card naming what is being asked instead of a bare prompt.

The e2e drain reduces to waiting for the queue to empty, and the legacy-link test accepts both correct outcomes of the create/pull race (consent-card assertion, or sealing straight to the pulled key).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5035,6 +5035,7 @@ version = "0.6.8"
 dependencies = [
  "aes-gcm",
  "anyhow",
+ "base64",
  "blake3",
  "bs58",
  "dialog-common",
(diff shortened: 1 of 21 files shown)


Gozala:
Folded into #773 (merged with checks bypassed after full local verification: e2e 20/20, wasm and native suites green on e100cded).

PR #776, opened 2026-08-26
refactor: retire POST /api/spaces

Gozala:
`POST /api/spaces` existed so the account gate's replay could create a space from `/account`, a page the original transient could not reach. That gate and its replay are gone from staging, and nothing in the app has called the endpoint since — only tests did.

Creating a space is a command. The `space/create` transient rides `/transact` like everything else, the handler navigates the originating client itself, and the outcome lands as facts the page already subscribes to. A second creation path reachable by one caller over one transport is what `route_table.rs` pins against, so the route, its pin entry, `post_space`, and both DTOs come out together.

The claim builder moves from `tonk-fab` to `tonk-worker-api` — the crate the FAB and `tonk-ui` already share — so every dispatcher builds the same transient from one definition.

## Two behaviours the endpoint was hiding

Converting the tests surfaced real differences between the two paths, both consequences of a command answering with facts and page effects rather than a response body. They are worth knowing about beyond the tests:

**The remote attaches asynchronously.** The endpoint attached before responding, so a caller could create-with-remote and push immediately. The handler navigates first and attaches after — deliberately, so navigation does not wait on the network — so that same push now races the attach and answers `SYNC_UNAVAILABLE`. The test helper waits for the attach when a remote was requested, so callers do not each have to know.

**Creating a space navigates into it.** The handler posts a navigate effect to the originating client, so the driver ends up on `/space/{key}`. A test that then read the account panel was on the wrong page.

## Verification

Full real-browser suite 20/20 against a live access service, worker wasm 306, fab 47, native workspace green, clippy `--all --all-targets --all-features -D warnings` clean.

Every failure seen along the way was baselined against clean staging before being attributed — two were mine (above), one was a stale `tonk-cli` binary in the harness.

---

## PR-Codex overview
This PR focuses on refining the handling of remote spaces in the `tonk` application, particularly around account registration states, space creation, and remote attachment logic. It introduces new constants for account-related refusals and updates various functions to streamline the space creation process.

### Detailed summary
- Removed `CreateSpaceRequest` and `CreateSpaceResponse` structs from `identity.rs`.
- Added new constants in `share.rs` for account-related refusal states: `BLOCKED_NEEDS_ACCOUNT`, `BLOCKED_NEEDS_ACTIVATION`, and `BLOCKED_SUSPENDED`.
- Updated the `explain_refusal` function to handle new refusal states based on registration status.
- Refined the `put_shortcut` function to clarify remote attachment requirements.
- Simplified `create_space_claim_json` usage across modules.
- Updated tests to reflect changes in refusal handling and space creation logic.
- Adjusted routes in `route_table.rs` and `router.rs` for API endpoints.
- Enhanced integration tests to verify new behavior around space creation and remote attachment.


Gozala:
Folded into #783, which now targets staging directly. Same commits, one PR instead of a stack.

PR #803, opened 2026-08-27
fix(register): tell the account panel the ceremony gave it an account

Gozala:
Twenty-six of the forty-three end-to-end account tests on #783 fail, and twenty-four of them fail on the same line inside `sign_up`:

```
Error: timed out waiting for `tonk-account[data-mode="success"]`
```

That is one culprit wearing twenty-four masks, and two more behind it.

## The panel is never told the account exists

`530d0d8` moved the signup and login ceremonies out of the `tonk-account` panel and into the registration cluster. The panel still derives its mode exactly twice — in `connected_callback`, and on `popstate` — and the cluster does neither: `close()` removes its own node and nothing else.

So the panel loads with no account, lands on `choice`, the cluster creates the account, the cluster closes, and the panel is still on `choice`. Not only in tests: someone who links an account from `/settings` is left looking at "link an account" for the account they just made. It is also why `it_registers_before_linking_a_cli_from_a_fresh_browser` never reached `handoff` — the comment on that path says "once the ceremony settles this reloads into the approval panel", and nothing was reloading it.

The cluster now says so, with one event on the top page the moment the ceremony settles. The panel re-reads the status rather than being handed a state to render, the same way it already re-derives on `popstate`. Announced at the ceremony rather than on close, because the cluster can settle with the dialog still up waiting on an emailed link.

## An answer the dialog already had, thrown away

`it_refreshes_the_address_answer_when_the_account_activates` failed with `the cluster never offered "create a passkey"; it shows ""`.

It asks about the address *before* registering, so the `EmailStatus` row already reads `{activates@example.com, unregistered}` when the dialog opens. Two things then conspire:

1. The subscription's snapshot delivers that row before anything is typed, and `show_answer` drops it as an answer about nothing on screen.
2. Typing re-asks. The handler writes `checking`, then writes `unregistered` again. Polls coalesce, and an established subscriber is sent a frame only when the poll reported a change — so the net change against what was already delivered is nothing, and no frame arrives at all.

The dialog was waiting for a frame that could not come. Every other test escapes it because the overlay held no prior row, so their first write is a real change.

It holds the newest answer now and renders it against what is typed. `show_answer` still decides whether the answer is about the current address, so nothing stale leaks through — it is the same single source of truth, just not discarded.

## Two things the test had wrong

`it_creates_a_local_only_spot_from_the_hub_wizard` reported `no invite link. profile main answered: {"body":[],"status":200}`.

- `await_share_link` asked profile main for `xyz.tonk.credential/link`. The mint writes `xyz.tonk.invite/url` on the **space's** branch — where the dialog itself reads it to fill the clipboard. `[]` was a correct answer to the wrong question.
- The test expected the link without taking the steps that produce one. The closing action appears once the address settles as verified and a display name is committed, and the link exists only after it is clicked. Its sibling `it_signs_up_to_share_and_hands_over_the_link` walks all of it and passes.

It also activated in the tab holding the ceremony. The cluster is a DOM element with no persistence, so navigating away and back destroys it and the confirmation comes home to nothing — the sibling test's second tab is what the emailed link opens anyway, so that is a helper now rather than a detail one test happens to know.

Signing in also waits for its passkey row before taking the cluster down, the way signing up does: dismissing on the click alone races an assertion the platform is still holding.

## Verification

`cargo check -p tonk-ui --target wasm32-unknown-unknown` and `cargo check -p tonk-ui --all-targets --features integration-tests` both clean; `cargo fmt --all` applied. The real-browser suite needs `nix run .#tonk-ui-test-server`, which this environment has no nix for, so the e2e run itself is CI's.

https://claude.ai/code/session_01B2QH6hFcNAxaPU9bFYyFi3

---
_Generated by [Claude Code](https://claude.ai/code/session_01B2QH6hFcNAxaPU9bFYyFi3)_

---

## PR-Codex overview
This PR focuses on enhancing the registration and account activation processes within the application, improving user experience, and ensuring accurate state management across different components.

### Detailed summary
- Added new claims for account settings and address resolution in `data.json` and `data.js`.
- Updated `unrun` count in JSON files.
- Improved email status handling and account state management in `customer.rs` and `email_status.rs`.
- Enhanced registration dialog behavior in `register_dialog.rs`.
- Implemented activation confirmation in a new tab in `account_flow.rs`.
- Adjusted share link retrieval logic in `account_flow.rs` and `account.rs`.


Gozala:
`Tests (native, debug)` failed on `f5d39d9`, and it is not this branch's:

```
FAIL tonk-cli account::tests::it_resumes_the_exact_pending_attachment_after_reopen
panicked at rust/tonk-cli/src/account.rs:1838:10:
pending recovery must not wait for a browser: Elapsed(())
```

Two things put it outside this change. `tonk-cli` depends on `tonk-worker-api` only — not `tonk-worker`, not `tonk-ui` — so nothing here reaches it. And the failure is a bare timeout, not an assertion about behaviour: `rust/tonk-cli/src/account.rs:1832` bounds `link_with_operator` at one second, over a call that reopens the store and derives keys.

Run five times on an idle machine it passes every time, but the spread is the point:

```
0.28s   2.49s   0.08s   0.03s   0.07s
```

One run in five already spends 2.5× the budget with nothing competing for the box. In CI this job shares a runner with four other test matrices, so the bound is thinner than the work it is timing. It passed on `staging` and on this branch's previous head; it is variance, not a regression.

The assertion worth keeping is "recovery does not wait for a browser" — that is what `assert_no_browser_announcement` on the next line already proves, and it holds whether the call takes 30ms or three seconds. The timeout only needs to be short enough to catch a genuine block on stdin:

```diff
-            Duration::from_secs(1),
+            // Bounds a wait on a browser that never comes, not the work
+            // itself: reopening the store and deriving keys is ~2.5s at
+            // the tail on a loaded runner. `assert_no_browser_announcement`
+            // below is what proves nothing waited.
+            Duration::from_secs(30),
```

Leaving that to `tonk-cli`'s owner rather than widening a PR about the account-flow suite. Happy to push it here if you would rather it rode along.

For the record on what this branch is for: end-to-end went from 17 passed / 26 failed to 41 passed / 2 failed, and the two that remained on `f5d39d9` were a regression from my own previous commit — a second `GET /api/customer` probe racing the one the activation notice already makes, which duplicated the deferred backup replay. `7fd5072` removes it.

---

Gozala:
E2E is green on `7fd5072`: **43 passed, 0 failed** (221s), from 17 passed / 26 failed (926s) on the base.

`Tests (native, debug)` also passed on this head, which settles the `tonk-cli` `it_resumes_the_exact_pending_attachment_after_reopen` timeout I raised earlier — same code, unloaded runner. The 1s bound over ~2.5s of tail work is still tight, so the proposed patch in that comment stands on its own merits; it is no longer blocking anything here.

### The four defects

Nearly all 26 failures were one bug wearing 24 masks.

1. **The settings panel was never told the account exists.** Moving the ceremony into the registration cluster left `tonk-account` deriving its mode only on connect and on `popstate` — neither of which fires when the cluster creates an account. A person who linked an account from `/settings` was left looking at "link an account". The cluster now announces `tonk:account-changed`; the panel re-derives, as it already did for history.

2. **A repeat answer produced no frame.** `EmailStatus` is a singleton row, so re-asking about an address it already names is a write that changes nothing, and an established subscriber is sent nothing when the delta is empty. The dialog was waiting on a frame that could not arrive. It now holds the newest answer and re-renders it against what is typed.

3. **`record_customer_status` never refreshed the address answer.** `register_dialog.rs` said it did and the worker test said `republish` mapped the statuses — but no such code existed, and the unit test passed vacuously because it only asserted `state_for_status`. An address asked about before registering kept `unregistered` forever.

4. **Two tests asked the wrong question.** `await_share_link` queried profile main for a row the mint writes on the space's branch, and the hub-wizard test expected a link before clicking the step that mints it, while activating in the ceremony's own tab and destroying the cluster.

Only #4 was the tests' fault; the rest were real defects the suite was correctly reporting. `LIFE-25` and `LIFE-26` in `verification/accounts.md` record #1 and #2 as claims a run can fail.

One regression along the way is worth flagging for reviewers, since it is a sharp edge in this code: my first fix for the unhydrated message called `activation_pending()` on the panel's load path, which is a second concurrent `GET /api/customer`. That endpoint is not a read — it holds a state guard and runs `drain_pending`, the deferred backup replay — so two in flight during the activation window duplicated the replay and `data-backup` never settled. The load now marks `data-unhydrated-on-load` and `render_registration` names the reason after the single probe.

---

Gozala:
`Tests (native, release)` failed on `7fd5072`, on a test this PR does not touch. Re-running the failed jobs once; recording the evidence here so it isn't re-derived.

```
tonk-cli::site when_migrating_from_carry::it_removes_the_source_when_moving
Error: verification failed; rolled back .../.tonk
  0: failed to save the native signing session
  1: Storage error: Authorization could not be evaluated: failed to retain delegation:
     Failed to publish new revision: Storage error: Lock error:
     Concurrent write in progress (lock held by pid 55515)
```

Not this PR's, on four independent grounds:

- **Unreachable from the diff.** Changed files are `tonk-ui` (3), `tonk-worker` (2), `docs/storybook` (3). `tonk-cli` depends on `tonk-worker-api` only — not `tonk-worker`, not `tonk-ui`.
- **The debug variant passed on this same head**, running these same `tonk-cli` tests.
- **E2E is green on this head** (43/43).
- **Does not reproduce locally.** 6× `cargo nextest run --release -p tonk-cli --test site` — 288 executions, 48/48 each time — plus a 48/48 `cargo test --release`. Run under nextest specifically, since it gives each test its own process, which is the only way the cross-process lock collision above can happen at all.

No fix ported, because there is nothing to port: the failure is contention, not a defect introduced anywhere I can see.

That said, it points at something real and pre-existing, offered as an observation rather than a change I'm making here. The test allocates its own `tempfile::tempdir()`, so the collision is not over the site directory — it is over the *native signing session* saved at `tonk-cli/src/site.rs:1101`, which lives outside the per-test temp dir. Any two `tonk-cli` tests that mint one can collide under nextest's process-per-test model, and the window is small enough that it only shows up on a loaded runner. This is the second distinct `tonk-cli` test to fail this way on this PR (the earlier one being the 1s timeout in `it_resumes_the_exact_pending_attachment_after_reopen`), which suggests the isolation gap rather than either test. Fixing it belongs in a PR that owns `tonk-cli`.

For the record on the earlier one: `Tests (native, debug)` passed on this head, so that timeout was the same story — same code, unloaded runner.

---

Gozala:
Correction to the comment above: I could **not** re-run the job. Both `rerun-failed-jobs` and `rerun` return `403 Resource not accessible by integration` — this app lacks the permission. The evidence in that comment stands; only the "re-running once" line was wrong.

So the re-run is yours to spend if you want it. I'm not making `tonk-cli`'s signing-session isolation robust from here — that is a different crate from the ones this PR changes, and widening it to chase a failure I can't reproduce would be the wrong trade.

I'll keep watching until this is green, merged, or closed.

---

PR #863, opened 2026-09-02
fix(account): leave the deletion result on a fresh settings page

Gozala:
Account deletion looked broken from the settings page. The purge, unlink, and profile rotation all completed, but the "back to settings" action on the result was a client-side route to `/settings`, which `navigate_to` treats as a no-op from `/settings`. The "complete" dialog stayed open over the deleted account's panel, still showing its email, passkey, and sign-out controls.

The result action now loads the settings document afresh, the way sign-out reloads, so the page rebuilds on the rotated profile. That also drops any `?delete-space` entry a CLI-opened space deletion carried, which would otherwise re-open the flow for a space that is gone.

The e2e test posted to the delete API directly and never saw the dialog. It now drives the real review, typed address, acknowledgement, passkey gesture, and the result action, and asserts the page lands in the fresh profile's mode before re-signing up on the released email.

---

## PR-Codex overview
This PR focuses on improving the user experience during account deletion by ensuring proper navigation and updating the UI to reflect the changes after a profile is deleted.

### Detailed summary
- Added comments explaining the behavior after permanent deletion.
- Changed navigation method to ensure the settings document is loaded afresh.
- Updated the account deletion flow with additional UI interactions and checks.
- Ensured that the result dialog does not persist after leaving the deletion confirmation.

### DIFF
--- a/rust/tonk-ui/src/account.rs
+++ b/rust/tonk-ui/src/account.rs
@@ -3047,8 +3047,16 @@ fn bind(host: &HtmlElement) {
         }
     });
 
+    // The result stands on a page that no longer describes the profile
+    // under it: permanent deletion rotated the browser onto a fresh
+    // profile, and a space deletion removed a listed space. A client-side
+    // route to the same path is a no-op, so load the settings document
+    // afresh, dropping any `?delete-space` entry that would re-open the
+    // flow for a space that is now gone.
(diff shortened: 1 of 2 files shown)


Gozala:
Superseded by #864, which deletes the account panel this fixed and carries the same e2e coverage through the hub settings page.

PR #884, opened 2026-09-03
fix(observability): restore the signup journey's open and settle events

Gozala:
The ordered-journey capture has been failing since #864 — masked by the workspace-cleanup teardown race until #880 fixed it (staging's own #864 run fails the same test). It caught two real telemetry gaps rather than a stale expectation:

- `open_registration` was only recorded by the bare `register_dialog::open()`, but the hub raises the dialog through the sealed-guest entry (`open_with_return_focus`), so the journey opened with no opening event. The record now lives in the shared funnel, after the already-open guard so a re-show never double-records.
- The `settle_account` pair (activation converging into a working account) was recorded by the `/account` dashboard that #864 retired, and nothing inherited it. The awaiting-confirmation branches mark the pending settle again, and `finish_ceremony` — where the activation now converges via the facts subscription — consumes the marker and records the pair, keeping the journey's terminal event on the page that actually finished.

No test changes: the capture's expected sequence stands as the schema of record.

---

## PR-Codex overview
This PR introduces functionality for handling account settlements in a web environment, specifically targeting the `wasm32` architecture. It adds mechanisms to track and manage account activation processes and to ensure that pending settlements are properly recorded and handled.

### Detailed summary
- Added `spawn_settle_probe` function to initiate account settlement.
- Introduced `mark_settle_pending` and `take_settle_pending` functions for session storage management.
- Updated `run_signup_ceremony` to call `mark_settle_pending`.
- Enhanced `finish_ceremony` to check for pending settlements and handle them accordingly.
- Added comments for clarity on the new behavior.
- Implemented asynchronous polling for account status during settlement.

### DIFF
--- a/rust/tonk-ui/src/account_observability.rs
+++ b/rust/tonk-ui/src/account_observability.rs
@@ -66,6 +66,60 @@ pub(crate) fn take_settle_pending() -> bool {
         })
 }
 
+/// Consume a pending settle on page load. A ceremony parked on the
+/// emailed link is often navigated away — the activation link opens in
+/// this very tab — so the dialog's own convergence never runs; the
+/// marker rides sessionStorage across the detour, and this probe
+/// finishes the journey the way the account dashboard's load once did.
+/// The `/api/customer` read is the probe that also REPLAYS the deferred
+/// account work, so polling it is what settles the account, not merely
+/// what observes it.
+#[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+pub fn spawn_settle_probe() {
+    if !take_settle_pending() {
+        return;
+    }
+    wasm_bindgen_futures::spawn_local(async move {
+        let mut attempt = WebAccountAttempt::start(
+            AccountAction::SettleAccount,
+            Surface::Settings,
+            Trigger::Recovery,
+            AccountState::PendingActivation,
+        );
+        for _ in 0..45 {
+            let active = crate::api::customer_state()
+                .await
+                .ok()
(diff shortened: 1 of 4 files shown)


Gozala:
Closing this — the `open_registration` half is already landed, and re-applying it would double-count.

`open_with_return` already records `open_registration` unconditionally, and did so at this PR's own base `b3b0ece`; the record came in with #857. This PR inserts a second `record_instant_success(AccountAction::OpenRegistration, …)` into the same function just after the open guard, and deletes nothing, so every raise of the dialog would be recorded twice.

Two other things that argue against merging as-is: the branch is conflicted against `staging`, and its base is behind #887, #888 and #891, which rewrote `register_dialog.rs` and `account_flow.rs` substantially.

**The `settle_account` half is still a real gap, and closing this does not fix it.** Nothing in the ceremony flow emits that event on `staging` — the action exists in `tonk-analytics`' enum and is documented in `docs/telemetry.md`, but no code path records it, and hasn't since #864 retired the `/account` dashboard that used to. The `finish_ceremony` and `run_signup_ceremony` hook points this PR used both still exist, so the settle work is re-cuttable as a small PR off current `staging` if that signal is wanted.

Worth correcting the premise for the record: the ordered-journey capture is not failing on this. Its expected sequence is `open_registration → create_account ×4 → activate_account ×2`, with no `settle_account` in it — at this PR's base or on current `staging` — so the missing settle never failed that test, and this PR would not have changed any red job.

---

PR #893, opened 2026-09-04
fix(ci): get staging's Test workflow green again

Gozala:
## What changed

`staging`'s `Test` workflow has been red for days across two unrelated causes. This addresses both.

**1. `Service-worker lifecycle tests` — a real bug in the test.**

The job fails on an assertion, not a timeout:

```
test at rust/tonk-ui/tests/service-worker.test.mjs:765:3
✖ a claimed first page stays ready while its offline generation fills
  AssertionError [ERR_ASSERTION]: claim starts the background fill
  false !== true
```

The test sends `{type:"claim"}`, spins a fixed **20 event-loop turns** waiting for the offline fill to request the shell, then asserts it has. But `claim` only reaches that request after `ensureOfflineGeneration()` awaits `readGenerationMarker()` and digests the manifest. That is real work; `setImmediate` turns are not. Idle, the fill arrives in 2 turns and the budget looks generous; on a contended runner the 20 turns drain while the digest is still outstanding.

Now it waits on the shell request itself, bounded by a generous ceiling so a fill that never starts still fails with the same message. Both assertions keep their meaning — they are still checked while the shell response is withheld. The service worker is unchanged; its behaviour was never wrong.

**2. `End-to-end shard 1..3` — a load-flaky suite.**

The shards fail on a different test each run — account ceremonies and service-worker upgrades alike — always by a deadline expiring, never by a wrong value. The same commit has passed every shard in one run and failed all three in another. `test-threads = 1` already stops the tests fighting each other, so the contention is the runner itself, which is 2–3× slower than the deadlines assume.

Those deadlines are ~40 hand-rolled literals scattered through `account_flow.rs`. Widening them one at a time would be guesswork against a suite that cannot be run outside CI, so the e2e nextest profile retries instead.

Nothing is skipped, disabled or quarantined. Every test still runs, and one that fails all three attempts still fails the run. The shard job allows 45 minutes against ~8 minutes of tests.

This second part is mitigation, not a cure. The durable fix is to make the ceremonies faster, or to give those deadlines one scalable knob so a slow runner is accommodated deliberately rather than by attrition.

## Verification

The service-worker fix is verified directly — the job **passes** on this branch, having failed on `staging`.

Reproducing it locally needed the missing ingredient, load. Saturating the CPU reproduces CI exactly:

```console
$ node --test --test-name-pattern 'a claimed first page stays ready' rust/tonk-ui/tests/service-worker.test.mjs
        claim starts the background fill
# pass 0
# fail 1
```

After the change, idle and under that same load:

```console
$ node --test rust/tonk-ui/tests/service-worker.test.mjs      # idle
# pass 67
# fail 0

$ # under nproc*3 busy loops, three consecutive runs
loaded run 1: # pass 67 # fail 0
loaded run 2: # pass 67 # fail 0
loaded run 3: # pass 67 # fail 0
```

The assertion is not hollow — with `extendOfflineGeneration` removed from the claim path so the fill never starts, it still catches it:

```console
    not ok 3 - a claimed first page stays ready while its offline generation fills
```

`node --check` passes; `.config/nextest.toml` parses.

The retry change can only be judged by CI, which is the point of it — it is aimed at a failure mode that does not exist on an idle machine.

## Storybook impact

Choose one and explain it:

- [x] No user-visible contract changed because: this changes how a test waits and how CI retries a flaky suite. No product code is touched, so no screen, journey, verification or triage ID is affected.

https://claude.ai/code/session_011GL7ZLoBMGKy1yJS1A6tQs

---

## PR-Codex overview
This PR enhances test reliability by introducing a retry mechanism and improving timeout handling in the `service-worker.test.mjs` file, ensuring that tests are less likely to fail due to timing issues.

### Detailed summary
- Added `retries = 2` in `.config/nextest.toml` for test retries.
- Introduced `FILL_START_TIMEOUT_MS` constant for fill request timeout.
- Changed `let shellRequested` to `let announceShellRequest` and updated its handling.
- Used a promise to manage shell request timing with `fillStarted`.
- Updated assertions to reflect changes in shell request handling.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -16,6 +16,15 @@ test-threads = 1
 # Report every failure: a fail-fast pass over an hour-long suite hides
 # everything after the first broken test.
 fail-fast = false
+# Each test drives a real Caddy + Chrome + wasm stack, and a shared CI
+# runner is 2-3x slower than the hand-rolled deadlines in these tests
+# assume: the same commit has passed every shard in one run and failed
+# three in another, tripping a different deadline each time. Retry the
+# flakes rather than widen forty scattered budgets by guesswork. This
+# disables nothing — a test that fails all three attempts still fails
+# the run — and the durable fix is to make the ceremonies faster or to
+# give the deadlines one scalable knob.
+retries = 2
 # The #852 quarantine is drained: every account flow it named is
 # repaired and back in the suite. To quarantine a newly broken test,
 # add a `default-filter` here excluding it, with a linked issue.
--- a/rust/tonk-ui/tests/service-worker.test.mjs
+++ b/rust/tonk-ui/tests/service-worker.test.mjs
@@ -20,6 +20,10 @@ import { dirname, join } from "node:path";
 const HERE = dirname(fileURLToPath(import.meta.url));
 const SW_PATH = join(HERE, "..", "assets", "service_worker.js");
 const CACHE_RS_PATH = join(HERE, "..", "..", "tonk-worker", "src", "cache.rs");
+// Ceiling for "the background fill got as far as requesting the shell". Only a
+// fill that never starts should reach it, so it is generous enough that a
+// contended runner cannot trip it on timing alone.
+const FILL_START_TIMEOUT_MS = 30_000;
 const source = readFileSync(SW_PATH, "utf8");
 
 /** A minimal Cache/CacheStorage good enough for the decisions here. */
@@ -774,8 +778,15 @@ describe("immutable generation install", () => {
     });
     const manifestHash = await sha256Hex(utf8(manifestText));
     let releaseShell;
-    let shellRequested = false;
+    let announceShellRequest;
     const shellReady = new Promise((resolve) => { releaseShell = resolve; });
+    // Settles the moment the fill asks for the shell, so the assertion below
+    // waits on that event instead of a fixed number of event-loop turns. The
+    // fill reaches this request only after reading the generation marker and
+    // digesting the manifest; on a loaded runner a turn budget drains while
+    // that work is still outstanding, which failed the test for timing rather
+    // than for behaviour.
+    const shellRequested = new Promise((resolve) => { announceShellRequest = resolve; });
     const { self, caches } = withGlobals({
       registration: { active: null, waiting: null, installing: {}, addEventListener() {} },
       fetchImpl: async (input) => {
@@ -783,7 +794,7 @@ describe("immutable generation install", () => {
         if (path === "/worker_bg.wasm") return new Response(wasm, { status: 200 });
         if (path === "/asset-manifest.json") return new Response(manifestText, { status: 200 });
         if (path === "/") {
-          shellRequested = true;
+          announceShellRequest();
           await shellReady;
           return new Response(shellBody, { status: 200 });
         }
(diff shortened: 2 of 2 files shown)


Gozala:
## `Tests (web, debug)` is red on `a095dd4`, and it is not this PR's

```
FAIL [1.600s] (1238/1583) tonk-worker router::join::tests::it_installs_a_claim_without_copying_the_space
  panicked at rust/tonk-worker/src/router/join.rs:3413:9:
  installing a claim onto a 3000-fact space copied 43 nodes, against 45 for
  the space itself — the copy is scaling with the space, so the diff has lost its base
```

The three sibling `Tests` jobs (native debug, native release, web release) were cancelled by matrix fail-fast, so they have no verdict on this head.

### Why it is not this PR's

This branch adds exactly two things on top of staging `3f43e3f`: an event-driven wait in `rust/tonk-ui/tests/service-worker.test.mjs` (a Node test file) and `retries = 2` under `[profile.e2e]` in `.config/nextest.toml`. Neither is reachable from a `tonk-worker` wasm unit test — the `e2e` profile is not the one this job runs under. `Tests (web, debug)` passed on staging `3f43e3f`, which is contained in this head, and it passed on both earlier heads of this branch.

### Why it flakes

The fixture already documents this failure mode, at `join.rs`:

> Tree shape is a pure function of its keys, and history keys carry the repository issuer. `put_repo` deliberately generates a fresh signer, which made this cost fixture sample a different tree on every run and occasionally turn one insert into a near-total rechunk. Pin the issuer so this test measures the install algorithm rather than key-distribution luck.

That pin is incomplete. It fixes the repository issuer to `[65u8; 32]`, but the identity that signs the commits being measured is still randomised per run, in `router.rs`:

```rust
fn session_nonce() -> u32 { *NONCE.get_or_init(rand::random::<u32>) }
let profile_name = format!("test-tonk-{}-{}", session_nonce(), SEQ.fetch_add(1, Ordering::Relaxed));
pub(crate) fn test_root_seed(profile_name: &str) -> [u8; 32] { /* XOR-folds the name into a seed */ }
```

So the root seed — and the session operator that performs the filler and claim commits — varies from run to run, and `SEQ` additionally varies with test ordering. Tree shape is still sampling. `based = 43` against `baseless = 45` is the near-total rechunk the comment names, not a broken install: the same source passed on staging minutes earlier.

This is a hypothesis, not a proof. The tests module is `#[cfg(all(test, target_arch = "wasm32", target_os = "unknown"))]`, so it cannot be run outside the browser harness and I could not reproduce it to confirm which key source is responsible.

### Proposed patch (not applied here)

Pin the remaining key source for this fixture only, rather than globally — `session_nonce` is deliberately random so that a runner reusing a browser profile does not collide across runs, and making it constant would weaken isolation for the whole worker suite. Give `claim_install_cost` a state whose profile name, and therefore whose root seed, is fixed, so the measured tree is a pure function of the fixture. If a residue of variance remains after that, the honest fix is to assert on the ratio the test actually cares about with a bound derived from a measured distribution, rather than a single `* 8`.

I have deliberately not applied this: it is unrelated to what this PR sets out to fix, it is in code this PR does not touch, and I cannot validate it locally.

### Re-run

I attempted to re-run the failed job and got `403 Resource not accessible by integration`, so I do not have the means. Someone with write access re-running `Tests (web, debug)` on this head should show it green. I am keeping this PR watched until it is.

### The rest of this head

| Job | Result |
| --- | --- |
| Lints | pass |
| Build end-to-end test archive | pass |
| Service-worker lifecycle tests | pass |
| End-to-end shard 1 / 2 / 3 | pass |
| End-to-end account tests | pass |
| Tests (web, debug) | **fail** (above) |
| Tests (native debug/release, web release) | cancelled by fail-fast |

The e2e retry did its job on this run, which is the first direct evidence it works — shard 2:

```
SLOW [> 60.000s] tonk-ui service_worker_upgrade::tests::it_reloads_every_update_aware_tab_after_controller_replacement
TRY 1 FAIL [ 107.622s] ...
TRY 2 PASS [  26.318s] ...
Summary [ 330.182s] 26 tests run: 26 passed (1 flaky), 52 skipped
```

Same test, 107s then 26s: the contention is transient within a job, so a retry lands in a fast window. Staging's own run on `3f43e3f` failed shards 2 and 3 on the same content, without retries.

---

PR #895, opened 2026-09-04
test(account): capture custody refusals in the ceremony diagnostic

Gozala:
Three account flows are failing on staging with the page parked on the wrong route, and every one arrives with a diagnostic that cannot be acted on: `diagnostic=""` and `consent={"awaitingChoice":false,"message":null,"present":false}`.

## Why that diagnostic says nothing

The consent card is removed immediately on success and four seconds after a failure (`custody_relay.rs:271-277`), so a sixty-second wait finds it gone either way. And the probe behind `diagnostic` only wrapped `console.error` matching `account LinkCli failed:`, while `custody_relay::report` warns rather than errors, under a `custody:` prefix — so every custody refusal was structurally invisible to it. The deletion flow had no probe installed at all.

That matters more than it sounds, because the reply protocol is unambiguous once you can see it: a worker reply carrying `error` **rejects** the page's promise (`rust/tonk-identity/src/install.rs:246-260`), and only a clean reply resolves. So the refusal text always exists somewhere — it was simply never captured.

## What this adds

Both console levels are watched and both prefixes matched, captures accumulate rather than overwrite, and the deletion flow gets the probe. The existing 48-character redaction still applies, so nothing credential-shaped reaches session storage.

The deletion timeout also now asks the worker whether the purge actually landed. `the page never landed on /` currently covers two failures with opposite fixes — the purge failed partway through one of `purge_inner`'s dozen fallible steps (including a network POST to `/ucan/`), or it succeeded and only the page's navigation was lost. A deletion plan that 404s means the account is already gone, which separates them in one line.

## Scope

This fixes none of the three tests. It makes their next failure name the failing step instead of the final URL, which is the prerequisite for fixing them without guessing — and it has already paid for itself: reading the card semantics carefully is what ruled out the `context_generation` fence from #891 as the cause, and prevented a wrong fix from landing.

Verified the injected script parses and behaves: warn and error both captured, unrelated lines ignored, long tokens redacted, multiple captures concatenated.

---

## PR-Codex overview
This PR enhances the error handling and logging for the account flow in the `tonk-ui` application, particularly focusing on capturing diagnostic information during the link installation and ceremony processes.

### Detailed summary
- Updated documentation for clarity on diagnostics.
- Enhanced `install_link_error_probe` to capture errors from both `account LinkCli` and `custody`.
- Introduced `ceremony_log` function to log ceremony transitions from the service worker.
- Modified error handling to include ceremony diagnostics in various contexts.

### DIFF
--- a/rust/tonk-ui/src/account_flow.rs
+++ b/rust/tonk-ui/src/account_flow.rs
@@ -2276,6 +2276,45 @@ mod tests {
         Ok(())
     }
 
+    /// The worker's own account of the ceremony, from the service worker
+    /// log ring `/api/health` exposes.
+    ///
+    /// `ceremony::report` writes each transition through `log!`, so this
+    /// carries `authorize-device: working|done|failed <detail>` even when
+    /// the page's console says nothing — which is exactly the case a
+    /// silent stall produces. `/api/health` is answered from the worker's
+    /// glue rather than its wasm, so it stays readable when the worker
+    /// itself cannot answer.
+    async fn ceremony_log(driver: &WebDriver) -> String {
+        if driver.enter_default_frame().await.is_err() {
+            return "could not enter the top document".to_owned();
+        }
+        let health = match get_json(driver, "/api/health").await {
+            Ok(health) => health,
+            Err(error) => return format!("health unreachable: {error}"),
+        };
+        let lines = health["body"]["log"]
+            .as_array()
+            .map(|entries| {
+                entries
+                    .iter()
+                    .filter_map(|entry| entry["message"].as_str())
+                    .filter(|message| {
+                        message.contains("ceremony")
+                            || message.contains("authorize-device")
+                            || message.contains("custody")
+                            || message.contains("delete-account")
+                    })
+                    .map(str::to_owned)
+                    .collect::<Vec<_>>()
+            })
+            .unwrap_or_default();
+        if lines.is_empty() {
+            return "no ceremony lines in the worker log".to_owned();
+        }
+        lines.join(" | ")
+    }
+
     async fn link_error_diagnostic(driver: &WebDriver) -> Option<String> {
         driver
             .execute(
@@ -2372,9 +2411,15 @@ mod tests {
             let url = driver.current_url().await?;
             let diagnostic = link_error_diagnostic(driver).await.unwrap_or_default();
             let consent = custody_consent_diagnostic(driver).await;
+            // The worker's own account of the ceremony. An empty page-side
+            // diagnostic means the relay reported no failure, which cannot
+            // by itself separate "the worker never answered with a
+            // navigate" from "the ceremony never started".
+            let ceremony = ceremony_log(driver).await;
+            driver.enter_default_frame().await?;
             let status = get_json(driver, "/api/account").await?;
             return Err(wait_error).context(format!(
-                "approval never came back at {url}; account={status}; diagnostic={diagnostic:?}; consent={consent}"
+                "approval never came back at {url}; account={status}; diagnostic={diagnostic:?}; consent={consent}; ceremony={ceremony:?}"
             ));
         }
 
@@ -5155,12 +5200,15 @@ mod tests {
             // that still loads means the ceremony failed before finishing.
             // Without this the two are indistinguishable, and they have
             // opposite fixes.
+            let ceremony = ceremony_log(&driver).await;
+            driver.enter_default_frame().await?;
             let plan = get_json(&driver, "/api/account/deletion/plan")
                 .await
                 .map(|plan| plan["status"].clone())
                 .unwrap_or(serde_json::Value::Null);
             return Err(error).context(format!(
-                "deletion consent={consent}; diagnostic={diagnostic:?}; planStatus={plan}"
+                "deletion consent={consent}; diagnostic={diagnostic:?}; \
+                 planStatus={plan}; ceremony={ceremony:?}"
             ));
         }
         enter_hub(&driver).await?;
--- a/rust/tonk-ui/src/account_flow.rs
+++ b/rust/tonk-ui/src/account_flow.rs
@@ -5150,8 +5150,17 @@ mod tests {
             // alone cannot say which happened. The probe's captured
             // refusal is what distinguishes them.
             let diagnostic = link_error_diagnostic(&driver).await.unwrap_or_default();
+            // Whether the purge itself landed: 404 means the account is
+            // gone and only the page's navigation was lost, while a plan
+            // that still loads means the ceremony failed before finishing.
+            // Without this the two are indistinguishable, and they have
+            // opposite fixes.
+            let plan = get_json(&driver, "/api/account/deletion/plan")
+                .await
+                .map(|plan| plan["status"].clone())
+                .unwrap_or(serde_json::Value::Null);
             return Err(error).context(format!(
-                "deletion consent={consent}; diagnostic={diagnostic:?}"
+                "deletion consent={consent}; diagnostic={diagnostic:?}; planStatus={plan}"
             ));
         }
         enter_hub(&driver).await?;
--- a/rust/tonk-ui/src/account_flow.rs
+++ b/rust/tonk-ui/src/account_flow.rs
@@ -2220,10 +2220,17 @@ mod tests {
 
     const LINK_ERROR_DIAGNOSTIC: &str = "tonk:test:link-error";
 
-    /// Preserve the account panel's exact LinkCli diagnostic across the
-    /// activation detour. The visible copy is deliberately safe/generic; this
-    /// test-only probe makes a failed real-browser run identify which async
-    /// boundary failed without changing production UI or logging broadly.
+    /// Preserve the exact ceremony diagnostic across a navigation. The
+    /// visible copy is deliberately safe/generic; this test-only probe makes
+    /// a failed real-browser run identify which async boundary failed without
+    /// changing production UI or logging broadly.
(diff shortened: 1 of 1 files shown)


Gozala:
First real data from this branch, and it is a **negative result on my own premise** — worth recording plainly.

`it_backs_up_a_claimed_space_for_another_account_device` failed on shard 1 of this PR's run with the probe installed, and still reported `diagnostic=""`:

```
Error: approval never came back at .../settings/link?audience=...&callback=...&name=e2e%20terminal;
account={"body":{"accountState":"ready",...,"status":"registered"},"status":200};
diagnostic=""; consent={"awaitingChoice":false,"message":null,"present":false}
```

(Also note this is a **fifth** distinct link test failing across runs — `it_registers_before_linking`, `it_links_without_being_told_the_account_service`, `it_links_the_cli_through_the_browser_callback`, and now this one. The rotation is the story.)

## What the empty diagnostic rules out

I checked the realms rather than assuming. `custody_relay::install()` runs in the **top document** (`rust/tonk-ui/src/bin/ui.rs:44`), which is the same realm the probe wraps — so a `custody:` warn from `report()` would have been captured. The probe installs via `Page.addScriptToEvaluateOnNewDocument` **before** the `goto` (`account_flow.rs:2348`), so it covers the document under test.

So `diagnostic=""` is not a probe failure. It means **`report()` was never called**, which means `mediation.finish()` did not return `Err`. Combined with `present:false` (the card was removed, which only happens on the `Ok` branch — the `Err` branch lingers 4s), the ceremony **reported success and simply did not navigate**.

That is the discriminating fact I could not get before this branch, and it points somewhere specific: `Mediation::finish` (`custody_relay.rs:441-455`) navigates only when it finds `navigate` inside the reply's `ok`, and otherwise returns `Ok(())` having done nothing at all. The worker's `AuthorizeDevice` arm is supposed to answer `navigate_reply(&target)` (`custody.rs:315-318`). Something in between is losing that key — or the ceremony never ran and the card was never raised in the first place, which `present:false` alone cannot distinguish.

## What this changes about the PR

The probe as written cannot separate those last two cases, so it is not finished. The next iteration should record, in the top document, whether `run_command_ceremony` was entered and what the raw reply object's top-level keys were — that separates "worker never sent `navigate`" from "page got it and did not act" from "ceremony never started". I would rather say the instrument is incomplete than let `diagnostic=""` read as "nothing went wrong".

Still worth landing as-is: it converted an unreadable failure into a narrowed one, and the `planStatus` probe on the deletion path is untouched by this limitation.

Gozala:
Follow-up while tracing what an `Ok`-but-no-navigate ceremony could mean. One candidate found, one retracted — recording both so the next person does not re-walk either.

## Retracted: there is no mount/wire race

I thought `run_command_ceremony` had a window between `mount_card()` putting `#tonk-custody-continue` in the DOM (`custody_relay.rs:160`) and `on_click` attaching its handler (`:267`), which the test's poll-and-click (`account_flow.rs:596-600`) could slip into. **It cannot.** Both run synchronously in the same JS task, so a WebDriver click cannot interleave, and the DOM is not observable to the driver until that task yields. Not a bug.

## Candidate: `BUSY` is a latch with no timeout

`BUSY` (`custody_relay.rs:30`) gates both custody entry points — `run_command_ceremony` at `:253` and the `EncryptionKey` arm at `:329` — and both **return silently** when it is already set:

```rust
if BUSY.with(|busy| busy.replace(true)) {
    return;
}
```

It is cleared in exactly three places: `remove_card()` (`:109`), and the two `mount_card()` failure paths (`:169`, `:257`). So it is cleared only when the card is actually removed, and the card is removed only on a resolved ceremony — Continue clicked and finished, Dismiss clicked, or the 4s post-failure timer.

A consent card that is raised and never resolved therefore latches `BUSY` true **for the life of the document**. Every later ceremony is then dropped at `:253`/`:329` with no card, no console output, and no worker-side trace — which is precisely the shape observed: `finish()` returns `Ok`, `report()` is never called, `diagnostic=""`, and nothing navigates.

I am flagging this as a candidate, not a diagnosis. What would confirm it is the `ceremony=` field added in `cbe8d205`: if the worker log shows `authorize-device: working` with no matching `done`/`failed`, the worker asked and the page never answered, which is this. If it shows `done` with a target, the worker answered and the page lost it, which is elsewhere.

Worth noting either way: a silent latch with no timeout and no logging is hostile to diagnosis regardless of whether it is causing this particular failure. A dropped ceremony should at minimum say so.

Gozala:
Sharpening the `BUSY` candidate: the latch's own documented invariant does not hold for the commands it now guards.

`BUSY` is declared with this rationale (`custody_relay.rs:26-30`):

```rust
/// One card at a time: a second request arriving while the card is
/// up is already answered by the save the first one performs.
static BUSY: Cell<bool> = const { Cell::new(false) };
```

That is true for the case it was written for — `WebAuthnKind::EncryptionKey`, where every request wants the same thing (save the derived key with the root), so the first card's save genuinely answers the second request.

`run_command_ceremony` was added later, by #864 (`7156e05a`, the account-deletion PR), and reused the same latch for **arbitrary commands**. Six distinct intents now share it (`custody_relay.rs:234-239`):

- `PurgeAccount` → delete the account
- `AuthorizeDevice` → link a CLI
- `AddPasskey` → seal under another passkey
- `Enroll`, `CreateAccount`, `Login`

A pending deletion card does **not** answer a later device authorization. The invariant is simply false for these, so the dedupe collapses unrelated ceremonies into one another: whichever arrives second is dropped at `:253` with no card, no console output, and no worker-side trace.

That matches the observed failure shape exactly — `finish()` returns `Ok`, `report()` never fires, `diagnostic=""`, nothing navigates — and it would explain why the failing link test **rotates** between runs rather than being one specific broken flow. Whichever ceremony happens to run second in a given shard ordering is the one that gets swallowed.

Still a candidate, not a diagnosis. The `ceremony=` field added in `cbe8d205` is the discriminator: `working` with no terminal `done`/`failed` means the worker asked and the page never answered, which is this. Independent of whether it causes this failure, a latch that silently drops a ceremony with no log is worth fixing on its own terms — and the fix should probably key on the intent rather than a single global flag.

Gozala:
Two findings from this branch's run that change the picture.

## 1. The deletion test is a fixed ~71s timeout, and it passed here in 11s

Its timing across every run I have logs for:

| run | result |
|---|---|
| staging 33810190618 | FAIL **71.081s** |
| staging 33852153624 | FAIL **71.963s** |
| PR #894 | FAIL **70.818s** |
| PR #896 | FAIL **70.702s** |
| **this branch** | **PASS 11.428s** |

Four failures within a 1.3-second band is not a variable race — that is something waiting out a fixed deadline. And the passing run is 6x faster, not marginally faster, which says the fast path and the stuck path are qualitatively different rather than load-dependent.

I want to be explicit that **this PR is diagnostics-only and cannot have fixed it.** The plausible readings are (a) it is genuinely intermittent at a rate the four-run sample happened to miss, or (b) something about ordering or timing on this branch avoids the trigger. Neither is established. What matters is that "consistently broken" was the wrong model — it can pass, so whatever is wrong is conditional, and the ~71s constant should identify which deadline is being hit.

## 2. A sixth rotating link test, with a different failure mode

Shard 3 failed on `it_authorizes_a_waiting_cli_from_the_browser`:

```
Error: the page never delivered an authorization
Caused by: deadline has elapsed
```

That brings the rotating set to six: `it_registers_before_linking`, `it_links_without_being_told_the_account_service`, `it_links_the_cli_through_the_browser_callback`, `it_backs_up_a_claimed_space_for_another_account_device`, and now this one — plus shards 1 and 2 going fully green on this same run, where they failed on staging.

Note this failure does **not** come through the instrumented helper, so it carries no `ceremony=` field. The probe covers `link_cli_with`'s approval wait; this test asserts on a different path. So the diagnostics in this PR are narrower than the failure set they need to cover, and that is worth fixing before relying on them.

Gozala:
Arithmetic on the deletion test pins where it fails, which narrows this considerably.

`await_url_path` waits 60s (`account_flow.rs:3632`). Subtracting that from each failing run:

| run | total | minus the 60s wait |
|---|---|---|
| staging 33810190618 | 71.081s | **11.08s** |
| staging 33852153624 | 71.963s | **11.96s** |
| PR #894 | 70.818s | **10.82s** |
| PR #896 | 70.702s | **10.70s** |
| this branch (PASS) | 11.428s | — |

The pre-wait work takes 10.7–12.0s on every failing run, and the **entire passing run** is 11.4s. So every run performs identical work up to `await_url_path`, and then the navigation either happens promptly or never happens at all. There is no middle — no run navigates slowly.

That rules out a slow ceremony, a loaded runner, and a timeout that is merely too tight. The purge work completes in about the same 11 seconds either way; what differs is solely whether `location.assign("/")` is reached.

Which puts the whole question on `Mediation::finish` (`custody_relay.rs:441-455`): it navigates only when it finds `navigate` inside the reply's `ok`, and otherwise returns `Ok(())` having done nothing at all. Binary outcome, binary cause. Either the worker's reply carried the `navigate` key or it did not.

The two candidates that survive this, both consistent with a binary split:
- the `BUSY` latch dropping the ceremony before it ever runs (#898), which would mean no reply at all; or
- `purge` returning `Ok` down a path that skips `navigate_reply("/")`.

The `ceremony=` field on this branch distinguishes them: `delete-account: working` with no terminal state means the ceremony was asked for and never answered; `done /` means the worker produced the target and the page dropped it. This run happened to pass, so it did not exercise the field — but the next failure will.

PR #896, opened 2026-09-04
fix(ui): two races that strand a tab on the outgoing service worker

Gozala:
`priorController` is read once, synchronously, when the page script evaluates (`index.html`). A tab that opens alongside an existing deployment can get there before the browser attaches its controller and capture `null` — and the `!priorController` branch then **returns** before the update check is ever armed.

That leaves the document deaf for the rest of its life, by three independent mechanisms:

- no `updatefound` listener, because the code that arms one is past the return;
- no `retire-if-superseded` nudge — and that one is unreachable twice over, since it also dereferences `priorController.postMessage` behind a guard requiring `controller === priorController`;
- its permanent `controllerchange` listener (`onControllerReplacement`) no-ops on its own `!priorController` guard. `cleanup()` is not the culprit here; it removes a *different* listener.

So the tab sits on the outgoing generation while its siblings move on, until someone reloads by hand. That is a user-facing staleness bug in its own right, not only a test artifact.

## Why this is the failing e2e test

`it_reloads_every_update_aware_tab_after_controller_replacement` opens a second tab with `new_tab()` + `goto`, which is exactly the race: whether that document evaluates before or after control attaches decides whether it stays update-aware. That explains the intermittency — the same test passes and fails across runs — and it fits the CI diagnostic rather than arguing with it. The stuck tab reported `controlled: true`, which looks like it rules out a null `priorController`, but does not: `priorController` is captured at evaluation while `controlled` is read much later by the probe.

## The fix, and what it actually changes

Adopt the controller the claim wins, then carry on into the update check.

**A correction to an earlier version of this description**, which claimed a first install still returns early. It does not. I traced the branch and both a cold first install and an unclaimed-joining-an-existing-deployment document now reach the fall-through with a non-null controller, so **both** proceed into the update check. The `if (!priorController) return` I kept only catches the case where control never arrives at all.

For a cold install that fall-through means one `registration.update()` and one extra `connectivity` message. `update()` against a registration whose worker is already the newest is a no-op that finds nothing, `watchIncoming(null)` yields `"unchanged"`, and the block cleans up — so the cost is one redundant update check on first load, not a behavior change. `a cold first install waits for activation, then claims` still passes unchanged. I would rather state this plainly than let the narrower claim stand.

## Verification

New test `a tab claimed after load still notices a later successor` drives the real `index.html` activation block through a `pageHarness` mode where an incumbent is active but this document is not yet controlled. It **fails without the fix** (no `retire-if-superseded` is ever sent) and passes with it. Full `test:sw` suite green with no regressions.

Independent of #894, which is hardening plus diagnostics on the same area and makes no claim to fix this.

---

## PR-Codex overview
This PR focuses on enhancing the update awareness of service workers in web applications. It modifies the handling of service worker controllers to ensure documents that are not initially controlled can still respond to updates effectively.

### Detailed summary
- Changed `const` to `let` for `priorController` to allow updates.
- Introduced `claimedThisLoad` to track document control status.
- Added checks to maintain update awareness for documents that are not initially controlled.
- Enhanced comment clarity to explain service worker behavior.
- Updated tests to verify behavior of unclaimed documents and their response to updates.

### DIFF
--- a/rust/tonk-ui/index.html
+++ b/rust/tonk-ui/index.html
@@ -356,7 +356,11 @@
                     );
                 }
 
-                const priorController = navigator.serviceWorker.controller;
+                // Not `const`: a document that evaluates before the browser
+                // attaches its controller adopts the one the claim below wins,
+                // so it stays update-aware instead of going deaf for its whole
+                // life.
+                let priorController = navigator.serviceWorker.controller;
                 let reloadRequested = false;
                 const onControllerReplacement = () => {
                     if (
@@ -436,10 +440,22 @@
                     try {
                         sessionStorage.removeItem(UPGRADE_RELOAD);
                     } catch {}
-                    observingInstallLifecycle = false;
                     await sendConnectivity();
                     markServiceWorkerReady();
-                    return;
+                    // A first install has the newest worker already and has
+                    // nothing to update to. An unclaimed document joining an
+                    // EXISTING deployment does: it adopts the controller the
+                    // claim just won and carries on into the update check, so
+                    // a later successor still reaches it. Returning here left
+                    // such a tab with no `updatefound` listener, no
+                    // `retire-if-superseded`, and a permanent
+                    // `controllerchange` listener disarmed by its own null
+                    // guard — deaf until the person reloaded by hand.
+                    priorController = navigator.serviceWorker.controller;
+                    if (!priorController) {
+                        observingInstallLifecycle = false;
+                        return;
+                    }
                 }
                 navigator.serviceWorker.removeEventListener(
                     "controllerchange",
--- a/rust/tonk-ui/index.html
+++ b/rust/tonk-ui/index.html
@@ -534,8 +534,18 @@
                     noteLife();
                     onWorkerState();
                 };
-                const onUpdateFound = () =>
+                const onUpdateFound = () => {
                     watchIncoming(registration.installing);
+                    // `registration.waiting` and the successor's
+                    // `statechange` land independently. A document that
+                    // heard "installed" before the registration exposed
+                    // `waiting` took no action, and `watchIncoming` returns
+                    // early for a worker it already watches, so nothing
+                    // re-checked — leaving the incumbent unwoken and the
+                    // successor waiting for good. Re-evaluate here, where
+                    // the registration has just changed.
+                    onWorkerState();
+                };
                 const onReplacement = () => {
                     if (
                         navigator.serviceWorker.controller &&
--- a/rust/tonk-ui/index.html
+++ b/rust/tonk-ui/index.html
@@ -440,20 +440,20 @@
                     try {
                         sessionStorage.removeItem(UPGRADE_RELOAD);
                     } catch {}
-                    await sendConnectivity();
-                    markServiceWorkerReady();
-                    // A first install has the newest worker already and has
-                    // nothing to update to. An unclaimed document joining an
-                    // EXISTING deployment does: it adopts the controller the
-                    // claim just won and carries on into the update check, so
-                    // a later successor still reaches it. Returning here left
-                    // such a tab with no `updatefound` listener, no
+                    // A document that never gained a controller has nothing
+                    // to be update-aware ABOUT, so it still stops here.
+                    // One that did adopts the controller the claim just won
+                    // and carries on into the update check, so a later
+                    // successor still reaches it. Returning unconditionally
+                    // left such a tab with no `updatefound` listener, no
                     // `retire-if-superseded`, and a permanent
                     // `controllerchange` listener disarmed by its own null
                     // guard — deaf until the person reloaded by hand.
                     priorController = navigator.serviceWorker.controller;
                     if (!priorController) {
                         observingInstallLifecycle = false;
+                        await sendConnectivity();
+                        markServiceWorkerReady();
                         return;
                     }
                 }
--- a/rust/tonk-ui/index.html
+++ b/rust/tonk-ui/index.html
@@ -361,6 +361,14 @@
                 // so it stays update-aware instead of going deaf for its whole
                 // life.
                 let priorController = navigator.serviceWorker.controller;
+                // Whether this document became controlled by the claim below
+                // rather than arriving already controlled. Such a document
+                // still needs update AWARENESS, but not a fresh update
+                // CHECK: `navigator.serviceWorker.ready` resolved against
+                // the current registration moments ago, so re-checking over
+                // the network buys nothing and costs a fetch that queues
+                // behind an incumbent pinned by its own open streams.
+                let claimedThisLoad = false;
                 let reloadRequested = false;
                 const onControllerReplacement = () => {
                     if (
@@ -456,6 +464,7 @@
                         markServiceWorkerReady();
                         return;
                     }
+                    claimedThisLoad = true;
                 }
                 navigator.serviceWorker.removeEventListener(
                     "controllerchange",
@@ -582,6 +591,15 @@
                 if (existingUpdate) {
                     watchIncoming(existingUpdate);
                     result = await outcome;
+                } else if (claimedThisLoad) {
+                    // Nothing is installing or waiting right now, and the
+                    // claim already resolved against the current
+                    // registration, so there is nothing to check and nothing
+                    // to await. Stay armed instead of tearing down: the
+                    // listeners registered above are this document's only
+                    // notice of a successor that installs later, and
+                    // `cleanup()` below would remove them.
+                    return;
                 } else {
                     try {
                         await registration.update();
--- a/rust/tonk-ui/tests/service-worker-claim.test.mjs
+++ b/rust/tonk-ui/tests/service-worker-claim.test.mjs
@@ -551,6 +551,30 @@ test("a tab claimed after load still notices a later successor", async () => {
   );
 });
 
+test("the retire nudge survives a successor that reaches installed before waiting", async () => {
+  // `onWorkerState` fires on the successor's statechange, and its nudge
+  // requires `registration.waiting === incoming`. The browser sets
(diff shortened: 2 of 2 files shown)


Gozala:
Scope note after seeing a second staging run finish, so this PR is not read as covering more than it does.

Current staging (`3f43e3fc`) fails four e2e tests, and the previous run failed four as well, with only partial overlap:

| test | run 33810190618 | run 33852153624 |
|---|---|---|
| `it_deletes_the_account_and_releases_its_email_and_profile` | fail | fail |
| `it_reloads_every_update_aware_tab_after_controller_replacement` | fail | fail |
| `it_registers_before_linking_a_cli_from_a_fresh_browser` | fail | pass |
| `it_links_without_being_told_the_account_service` | fail | pass |
| `it_links_the_cli_through_the_browser_callback` | pass | fail |
| `it_adopts_a_complete_second_generation_without_mixing_assets` | pass | fail |

Two things follow.

**The link tests rotate.** A different one fails each run, which is the signature of a flake rather than a break, and is why chasing any single one of them for a root cause was the wrong instinct.

**`it_adopts_a_complete_second_generation_without_mixing_assets` is NOT covered by this PR.** It produces a byte-identical stuck signature — `adopted` reached, `waiting: "installed"`, document pinned to the old build — but it uses a **single tab with `driver.refresh()`**, no second tab at all. A refreshed document should have its controller at evaluation time, so the null-`priorController` race this PR fixes does not explain it. It expects `documents: 3` and got 2, meaning the alignment reload never fired.

So there is at least one more failure mode in the successor-adoption path that this change does not address. I would rather flag that now than have this PR merge and the shard stay red for a reason nobody is looking for.

What this PR still stands on: the bug it fixes is real, reproduces deterministically as a unit test, and is a user-facing staleness bug independent of any e2e test.

Gozala:
Follow-up: the second failure mode I flagged above is now fixed here too, so this PR covers both stuck-handoff shapes rather than just the two-tab one.

**`it_adopts_a_complete_second_generation_without_mixing_assets`** is single-tab, so the null-`priorController` race could not explain it. The cause is a second, independent ordering race in the same block.

`onWorkerState` fires on the successor's `statechange`, and its `retire-if-superseded` nudge additionally requires `registration.waiting === incoming`. Those two land **independently** — the browser sets `waiting` and dispatches `statechange` without ordering them. A document that hears "installed" before the registration exposes `waiting` evaluates the nudge condition, finds it false, and does nothing. Nothing re-checks afterwards, because `watchIncoming` returns early for a worker it is already watching. The incumbent is never woken, the successor sits in `waiting`, and the tab keeps the outgoing build — with no controller replacement for the alignment reload to fire on. That is exactly why the test expects `documents: 3` and reports 2.

Existing coverage missed it because every test in this file sets `registration.waiting` **before** dispatching `statechange`, which is the favorable ordering the browser does not guarantee.

The fix re-evaluates on `updatefound`, where the registration has just changed.

New test `the retire nudge survives a successor that reaches installed before waiting` drives the adverse ordering. Verified against `origin/staging`'s actual `index.html`, both new tests fail there and pass with this branch:

```
not ok 9 - a tab claimed after load still notices a later successor
not ok 10 - the retire nudge survives a successor that reaches installed before waiting
```

Full suite 113 green, no regressions.

Both bugs strand a tab on a stale build in ordinary use, not only under test.

Gozala:
**Both fixes confirmed working in CI.** Same-shard comparison against the staging baseline (run `33852153624`, commit `3f43e3fc` — this branch's base):

| test | staging | this branch |
|---|---|---|
| `it_reloads_every_update_aware_tab_after_controller_replacement` | **FAIL 111.3s** | **PASS 28.9s** |
| `it_adopts_a_complete_second_generation_without_mixing_assets` | **FAIL 109.2s** | **PASS 24.2s** |

Both had been timing out at ~110s waiting for a handoff that never completed; both now pass in a quarter of that. The speedup is the tell — they are no longer waiting out a stall, which is exactly what the two races caused. Each fix has a unit test that fails against `origin/staging`'s `index.html` and passes here, so this is not a lucky run.

## One remaining failure in the same area, not fixed here

Shard 1 failed on a **third** test with a byte-identical signature:

```
FAIL [ 109.820s] it_releases_incumbent_streams_for_an_automatic_successor
Error: timed out waiting for coherent build a9aac2ed1ff649ea:
  {"active":"activated","waiting":"installed","documentBuild":"773ac09f71a7a2b1",
   "documents":2,...,"testInstallProgress":[...{"phase":"adopted","total":847}]}
```

`adopted` reached, `waiting: "installed"`, document pinned to the old build — the same shape as the two fixed here, so there is a third variant of this race that neither fix covers. Worth noting this test was already slow-but-passing on staging (86s, marked SLOW), so it was living just under the 90s budget and is not a new break.

I have not diagnosed it, and I am not going to guess at it in this PR. The diagnostic that would identify it — cache names plus the worker's log ring on the stuck path — is in #894, and this run's failure output still shows the bare `lifecycle` object with no `health` key, which is precisely the gap #894 closes. Landing #894 first would make the next occurrence self-explaining.

Recommendation: merge this on the two confirmed fixes; treat the third variant as follow-up work behind #894's diagnostics.

Gozala:
Net effect versus the staging baseline this branch is based on (`3f43e3fc`, run `33852153624`):

| | staging | this branch |
|---|---|---|
| failing e2e tests | **4** | **2** |

- `it_reloads_every_update_aware_tab_after_controller_replacement` — FAIL 111.3s → **PASS 28.9s**
- `it_adopts_a_complete_second_generation_without_mixing_assets` — FAIL 109.2s → **PASS 24.2s**
- `it_links_the_cli_through_the_browser_callback` — FAIL → pass (rotating flake, not addressed here; do not credit this branch)
- `it_deletes_the_account_and_releases_its_email_and_profile` — FAIL → FAIL (unchanged, not addressed here)
- `it_releases_incumbent_streams_for_an_automatic_successor` — PASS 86.1s → **FAIL 109.8s**

## On that last one, stated carefully

It passed at 86.1s on **both** staging runs, against a 90s budget — it was already 4 seconds from the cliff, and it carries the identical stuck signature (`adopted` reached, `waiting: "installed"`, document on the old build). So it is the third variant of the same latent race, and it was going to tip over on its own.

But I will not claim this branch is innocent of the extra 24s. This PR changes `index.html`, which is the page under test, and adds one `registration.update()` on the claimed path. That test opens a live SSE stream and never refreshes — it depends on a stream-pinned incumbent noticing a successor without restarting — so it is structurally sensitive to exactly the kind of extra work that path now does. I cannot rule out that the change pushed a marginal test over its budget, and the honest position is that this needs a rerun or two to separate "latent race finally tipping" from "made 24s worse".

If a rerun shows it consistently at ~110s on this branch and ~86s on staging, that is a regression and this PR should be held until the third race is fixed properly rather than merged on the two confirmed wins.

Gozala:
Measurement on the third variant, `it_releases_incumbent_streams_for_an_automatic_successor`, since it bears on whether this PR regressed it.

| run | result |
|---|---|
| staging 33810190618 | PASS **86.073s** |
| staging 33852153624 | PASS **86.119s** |
| this branch | FAIL **109.820s** |

**The staging spread is 46 milliseconds across two runs.** That is not runner variance — it is a deterministic wait landing on the same value twice. The test's only budget is `wait_for_mounted_build`'s 90s, so it was spending ~86s inside that wait and clearing it with under 4 seconds to spare, every time.

So this test has been stalling on staging all along; it simply stalled 4 seconds under the line instead of over it. It is the same family as the two fixed here — the incumbent holds two SSE readers open (`service_worker_upgrade.rs:1008-1016`), which is exactly the "stream-pinned worker never restarts" case the worker's own comments describe.

That does not clear this PR of the extra 24s. A deterministic ~86s wait plus a change on the claimed path is precisely the combination where a small added cost tips a marginal test. The rerun now in flight is the discriminator:

- ~86s again on this branch → the earlier 109.8s was noise on top of a latent stall, and this PR is clean.
- ~110s again → the change moved it, and this PR should wait for the third race to be fixed rather than merge on the two confirmed wins.

I have not diagnosed the underlying ~86s stall and am not going to guess at it here. It is a third variant of the same handoff race, it predates this branch, and #894's cache-name and log-ring diagnostics on the stuck path are the right instrument for it.

Gozala:
**The rerun says this PR regresses `it_releases_incumbent_streams_for_an_automatic_successor`. Do not merge as-is.**

| | run 1 | run 2 | spread |
|---|---|---|---|
| staging | PASS 86.073s | PASS 86.119s | **0.046s** |
| this branch | FAIL 109.820s | FAIL 109.650s | **0.170s** |

Both are deterministic — sub-200ms spreads on either side. This is not a marginal test bouncing across a threshold; it is a fixed wait that moved by **23.6 seconds** and stayed moved. The convenient reading I offered earlier ("it was 4 seconds from the cliff and was going to tip anyway") is wrong: it would predict scatter, and there is none.

So the change costs that path ~24 seconds. The two fixes are still real and still confirmed on both runs — `it_reloads_every_update_aware_tab_after_controller_replacement` passed at 28.860s and 28.811s, against 111.3s failing on staging — but they are not worth trading for a new deterministic failure.

## The likely mechanism

`it_releases_incumbent_streams` holds two SSE readers open and never refreshes, so its incumbent is stream-pinned and cannot restart. This PR makes a claimed document fall through into the update check, which runs `registration.update()` — a network fetch — on a path that previously returned early. Against a worker already pinned by open streams, that plausibly serializes behind them.

Fixing this properly means making the fall-through cheap for a document that has just been claimed, rather than having it repeat the full update check. That is a real design question, not a tweak, and I would rather reopen it than paper over a 24-second regression.

## Recommendation

Hold this PR. The two fixes want a narrower change on the claimed path — one that arms update awareness without paying for a fresh `registration.update()`. I will rework it rather than ask anyone to accept this trade.

PR #898, opened 2026-09-04
fix(ui): do not drop a command whose ceremony card is already up

Gozala:
**Status: a real defect, but an unverified fix.** I have not reproduced the e2e failure it is suspected of causing, and this module has no unit-test seam. Read it as "a silent drop that should not exist", not as "this turns the shard green."

## The defect

`BUSY` carries its own justification in the source (`custody_relay.rs:26-30`):

```rust
/// One card at a time: a second request arriving while the card is
/// up is already answered by the save the first one performs.
static BUSY: Cell<bool> = const { Cell::new(false) };
```

That is true for the case it was written for — `WebAuthnKind::EncryptionKey`, where every ask wants the same thing, so the standing card's save genuinely answers the second request.

`run_command_ceremony` was added later by #864 and reused the same latch for **arbitrary commands**. Six distinct intents now share it (`:234-239`): `PurgeAccount`, `AuthorizeDevice`, `AddPasskey`, `Enroll`, `CreateAccount`, `Login`. A pending deletion does not answer a later device authorization, so the documented invariant is false for every one of them.

The consequence is the part that matters: the second ceremony returns at `:253` with **no card, no console line, and nothing in the worker log**. Whatever asked for it waits until it times out, with no evidence anywhere of what happened.

## The change

Retire the standing card and run the new command, logging that it happened. The encryption-key path (`:340-344`) keeps its dedupe untouched, since its invariant genuinely holds.

## Why I suspect it in the e2e flakes, and why that is not a claim

The failing link test **rotates** between runs — `it_registers_before_linking`, `it_links_without_being_told_the_account_service`, `it_links_the_cli_through_the_browser_callback`, `it_backs_up_a_claimed_space_for_another_account_device` across four runs. That is not four broken flows; it is one mechanism surfacing wherever a second ceremony follows a first. A latch that swallows whichever ceremony arrives second would produce exactly that pattern, and exactly the observed shape: `finish()` returns `Ok`, `report()` never fires, `diagnostic=""`, nothing navigates.

But I have not confirmed it. `custody_relay` is `#[cfg(wasm32)]` and DOM-bound, so there is no native seam to write a failing test against, and I stood down the browser reproduction as too expensive for the return. The discriminator is already deployed in #895: a worker log showing `working` with no terminal `done`/`failed` means the page never answered the ask, which is this bug. If that lands and shows otherwise, this PR should be judged on its own merits rather than as a fix.

Independent of the flakes: a ceremony dropped with no diagnostic is worth removing regardless.

---

## PR-Codex overview
This PR enhances the `run_command_ceremony` function in `custody_relay.rs` by adding comments that clarify the behavior of handling commands and the `BUSY` state in relation to card management.

### Detailed summary
- Added comments explaining the purpose of the `BUSY` state in deduplication of identical asks.
- Clarified that a pending deletion does not respond to later device authorizations.
- Documented the need to retire the standing card when replacing it with a new command.


Gozala:
A concrete, fully-reachable path to this latch in the failing deletion test — which upgrades this PR from "a defect worth fixing" to a specific candidate cause.

## The sequence

1. `it_deletes_the_account_and_releases_its_email_and_profile` calls `create_space_awaiting_remote(..., true)` — a space with a remote (`account_flow.rs:5066`).
2. Provisioning that remote needs the account encryption key. When this device's root record predates that key, the worker asks the page for it: `request_and_wait` → `request_webauthn(client, WebAuthnKind::EncryptionKey)` (`custody.rs:90-96`).
3. The relay's `EncryptionKey` arm **takes `BUSY`** and raises a consent card (`custody_relay.rs:340-344`).
4. **The test never clicks it.** That card is worker-initiated; the test does not know it exists and has no step for it. `use_passkey_consent` is called only much later, for the deletion.
5. The worker waits `ASSERTION_TIMEOUT` = **120 seconds** (`custody.rs:31`) — longer than the entire test.
6. So the card is still up, and `BUSY` still true, when the deletion ceremony arrives ~11s later. `run_command_ceremony` returns at `:253`. No card, no console line, nothing in the worker log.
7. The page never navigates. `await_url_path` gives up at exactly 60s.

## Why this fits every measurement

The timing decomposition on #895 showed the pre-wait work takes 10.7–12.0s on every failing run while the entire passing run is 11.4s — identical work, then either an immediate navigation or none at all. A latch is exactly that shape: binary, with no slow middle. A race or a loaded runner would produce a spread.

It also explains the rotation across link tests. Whether the encryption-key ask fires at all depends on whether this device's root record predates the key — state that varies with test ordering and sharding — so the swallow lands on whichever ceremony happens to follow it.

## Caveats, because this is still inference

I have not reproduced it. Step 2 is conditional on the root-record check at `custody.rs:60-78`, and I have not confirmed that condition holds in this test's fixture. If the ask never fires, `BUSY` is never taken and this path is not the cause.

The `ceremony=` diagnostic in #895 settles it: `delete-account: working` with no terminal `done`/`failed` means the ceremony was asked for and never answered, which is this. It would also be worth logging the dropped-ask case directly — which this PR now does.

Gozala:
Correction to the confidence level of my previous comment, after checking the conditional step I flagged.

The encryption-key ask fires only when **both** hold (`custody.rs:55-78`):

```rust
if root.encryption_key.is_some() { return Ok(()); }
// ...
if published_sealed_inbox(&tonk, &root.root_did).await?.is_some() { return Ok(()); }
```

And the call site's own comment (`repository.rs:575-577`) names the case precisely: *"A linked device whose root record predates the encryption key asks the originating page for a passkey assertion here."*

The deletion test **signs up fresh** rather than linking a device with an older root record, so the case the comment describes is not obviously the one it is in. If enrollment populates `encryption_key` on the root — which a fresh signup plausibly does — then `ensure_recipient` returns early, no card is raised, `BUSY` is never taken, and **the chain in my previous comment does not apply to this test**.

I cannot settle that statically: it depends on what enrollment actually writes at runtime, and `local_root` reads a stored record rather than deriving it. So the previous comment should be read as "here is a reachable path", not "here is what happens in this test". I was more confident there than the evidence supports.

What survives unchanged, and is why this PR still stands on its own:

- The latch's documented invariant is false for the six command intents that now share it. That is a defect independent of any test.
- A dropped ceremony produces no card, no console line, and no worker-log entry. Whatever the cause of the e2e failures turns out to be, a silent drop is the wrong behavior and actively obstructs diagnosis.

The measurement that is not in doubt: the deletion test's pre-wait work takes 10.7-12.0s on every failing run and 11.4s on the one passing run, so the ceremony either navigates promptly or never. Something binary gates it. This latch is one mechanism with that shape; it is not yet shown to be the one.

Gozala:
Settled: the encryption-key path is **not** what breaks the deletion test. Ruling it out, with the evidence.

The crate's own unit tests answer the question I could not answer by reading the call site:

- `it_needs_no_assertion_when_the_root_carries_the_key` (`custody.rs:1265-1269`) — *"The fixture's root carries the key a ceremony would have recorded."* `ensure_recipient` returns without asking.
- `it_needs_no_assertion_without_a_passkey_root` (`:1258-1262`) — an onboarding device derives its recipient locally; nothing to ask.
- `it_refuses_without_a_client_to_ask` (`:1303-1310`) — the ask fires **only** after `strip_recorded_key`, a helper documented as *"Overwrite the local root record without its recipient and drop the published fact, the shape of a device linked before the key existed."* (`:1313-1316`)

So the ask requires a root record deliberately stripped of its key. `it_deletes_the_account_and_releases_its_email_and_profile` signs up fresh, so its root carries the key, `ensure_recipient` returns early at `custody.rs:67`, no consent card is raised, and `BUSY` is never taken before the deletion ceremony.

**The chain I described two comments ago does not occur in that test.** I am striking it rather than leaving it to be re-investigated.

## What this PR is now

A real defect with no demonstrated connection to any failing test:

- `BUSY`'s documented invariant ("a second request is already answered by the save the first one performs") is true for `EncryptionKey` and false for the six command intents `run_command_ceremony` later reused it for.
- A ceremony dropped by that latch leaves no card, no console line, and no worker-log entry.

That is worth fixing on its own terms — a silent drop obstructs exactly the kind of diagnosis this whole investigation has needed — but it should be reviewed as cleanup, not as a fix for the e2e failures. The binary gate on the deletion test's navigation is still unexplained, and I would rather leave it open than attach it to this.

PR #905, opened 2026-09-06
Resolve a view's event bindings at lowering

Gozala:
Stacked on #904 — review that first; this PR's diff is the commits on top of it. #906 is stacked on this one.

## What changed

A `show:` template binds an interaction with `on:<name>=<command>`. Both halves were resolved at **render** time: one query per declaration name to rebuild the event descriptor, one per command name to fetch its shape. A name that resolved to nothing produced no listener and no diagnostic — the element was simply inert, and the only way to find out was to click it.

Both halves are knowable at **lowering**: the analyzer has the template text and the document's `event!:` / `command!:` declarations in scope. So it now scans the templates, resolves each binding, fails the lowering on anything that does not resolve, and stores the result on the view as one dag-cbor artifact the display decodes — the same move `concept!:` already makes when it inlines its attributes' descriptors instead of leaving names for the runtime to chase.

The change proper:

1. **`view` becomes a built-in concept.** It has to be, for the analyzer to recognise a view structurally. Adds an optional `bindings` field at `xyz.tonk.view/bindings`.
2. **The libraries stop declaring `concept!: &view`.** An in-document declaration shadows a builtin, so leaving them in would silently keep runtime resolution.
3. **The inlining pass.**
   - `tonk-template::bindings` — a DOM-free scan for `on:` attributes (HTML comments and non-binding colon attributes excluded, matching the browser's DOM walk) plus the `Bindings` artifact. dag-cbor is canonical, so re-lowering an unchanged view yields byte-identical output and the claim does not churn.
   - `tonk-analyzer::view` — indexes the document's `event!:` declarations (resolving bare-symbol sources to entities once, so the event-time path never looks a name up), resolves every binding, and emits the `bindings` term.
   - `tonk-display` — reads the artifact once per refresh and skips per-name resolution for every declaration it covers.
4. **The field is a `Record`, and the payload identifies itself** — see below.

Then the fallout, each its own commit so the diff explains itself:

5. **The CLI site fixtures stop declaring a concept named `view`** — the fallout of (1), see **Compatibility**.
6. **The display stops holding its state borrow across the bindings query** — the one real defect this PR shipped, see below.
7. **Tests for the display half**, which had none.
8. **A `tonk-fab` settle budget raised to match its own file's other one** — see **Verification**.
9. **The new diagnostics point at the `on:` attribute rather than the whole template** — see below.

Three new diagnostics, all at build time instead of at click time:

| code | what it catches |
| --- | --- |
| `E_UNKNOWN_EVENT_DECLARATION` | `on:<name>` names a declaration nothing declares |
| `E_UNKNOWN_BOUND_COMMAND` | the command half resolves to no concept |
| `E_EVENT_COMMAND_MISMATCH` | the declaration cannot fill the command — a required field with no source, or a source naming a field the command does not declare |

The last one is the check the whole `event!:` split exists to make possible. It previously lived only in a `tonk-worker` test that string-scanned the shipped YAML; it now runs on every document the analyzer sees.

What is deliberately **not** inlined is the command's descriptor. The display already resolves a concept from a name for every model it renders, so a second copy would duplicate work rather than remove it — the analyzer's contribution for commands is the check.

### Where those diagnostics point

All three initially reported against the `show:` entry's whole value range. A template is a block scalar many lines wide, so an editor underlined the entire thing to say "one of these attributes is wrong" — when the analyzer knows exactly which eight characters are.

Commit (9) narrows them. `bindings::scan` now records each binding's byte offset within the template, and the analyzer maps that offset into the document's own coordinates: every content line of a block scalar starts at the value range's column, so the line offset adds to the line and the indentation adds to the column. A computed position that would fall outside the value — a multi-line flow scalar whose escapes shift the columns — falls back to the whole range, since a report that is merely wide beats one that points at the wrong text.

### What this does *not* check

Worth stating plainly, because it is easy to read the above as "command dispatch is verified now". It is not. These checks cover **notation**: `on:` bindings inside a `show:` template, resolved against the `event!:` / `command!:` declarations in scope when a document is lowered.

`tonk-fab` dispatches commands by a second, parallel path that no document describes. It hand-builds each claim as JSON in Rust and inlines its own descriptor, deliberately — `core.yaml` is seeded once at repo creation and frozen, so the FAB is built to consult nothing seeded (see the header of `fab_drift.rs`). Nothing the analyzer does can see those claims, because there is no document to lower.

That gap is not theoretical: renaming a space was broken on this stack the whole time, and #904 now carries the fix (`6e1a35d`). The descriptor half of the FAB's rename claim was migrated to the new namespace and the parameters half was not, so every rename failed with `invalid claim: field "rename-repository" is not declared by this concept` and the chip reverted with no other symptom. It was found by using the preview, not by any check here.

The durable fix is to stop hand-writing those claims: `tonk-fab` already depends on `tonk-schema`, and the command structs are `#[derive(Concept)]`, so both halves could be derived from the struct and could not disagree. That is a rewrite of every FAB claim builder and is not in this stack.

**`{field}` interpolations are unchecked *here*.** A `show:` template interpolating `{name}` where `name` is not a field of the view's model renders nothing, silently — the same class of failure this PR fixes for `on:` bindings, on the other half of the template. **#906, stacked on this PR, closes it**, along with the equivalent check on a bound declaration's `where:` sources. It is a separate PR because making it usable turns on getting three renderer exclusions right — `<style>` / `<script>` bodies, comments, and portal documents — and without them the check rejects the library it exists to protect. No shipped library has such a miss today; that PR catches the next one.

### The borrow this shipped, and what it cost

Commit (3) added an awaited query to `refresh_delegate`, reading the model entity like this:

```rust
let inlined = match state.borrow().model_entity.clone() {
    Some(model_entity) => resolve_inlined_bindings(host, &model_entity).await,
    None => BTreeMap::new(),
};
```

A `match` keeps its scrutinee's temporaries alive for the whole match, so the `Ref` was held across the await. Any view frame, entity frame, or disconnect landing in that window calls `state.borrow_mut()` and panics. The window is open only while that one query is in flight, so it failed by timing — which is why two browser jobs failed on one head and passed on another carrying the same code, and why "the artifacts are identical to a head where this passed" was the wrong thing to conclude from. Commit (6) snapshots the entity in its own statement, the shape the view-element snapshot ten lines above already uses.

Commit (7) pins it, along with the rest of the display path. Getting that regression test to actually fail on the old code took two attempts: asserting it by panicking inside the stub host proves nothing, because an exception raised in a DOM listener is swallowed by `dispatch_event` and only makes that query answer nothing, which the fallback treats as an ordinary miss. It now observes the borrow with `try_borrow_mut`, and tells the bindings query apart from the others by running the refresh twice — with and without a resolved model — and requiring exactly one more query, since the bodies arrive as JS `Map`s that `JSON.stringify` renders as `{}`. Verified in both directions.

### Why `Record`, and why the payload still tags itself

The field started as `Bytes`. That was the wrong type: dialog has `Record` for exactly this shape — one struct serialized as one atomic fact — and already uses it that way for `RevisionRecord`, whose read path matches on `Value::Record(bytes)` to tell it apart from an opaque blob. The type byte is carried in the key encoding, so the distinction survives storage and any `Value` match.

It survives nothing past that. `Record` says "structured", not *which* structure, and every wire projection flattens it into the shape a `Bytes` takes — `Ipld::Bytes` to the display, base64 in the claim API — while `Value`'s `#[serde(untagged)]` repr resolves any byte payload back to `Bytes` regardless of which variant produced it. So the artifact names its own format: `Bindings` carries `kind: "tonk/view-bindings@1"` and `decode` refuses anything else rather than quietly reading it as an empty binding table. Belt and braces, but the failure it prevents is the silent one.

`record` also joins the `as:` vocabulary (`text`, `bytes`, `entity`, …). The built-in `view` declares one, and a value type the compiler can write but an author cannot spell would be a schema only the compiler could author.

### Compatibility

- A binding whose declaration lives in an already-seeded document (a component library binding a `core.yaml` event) still resolves as a *name*, so it is not inlined and not an error. Only a name that resolves to nothing at all fails. A view carrying no `bindings` at all — one seeded before this existed — falls back to the old per-name path unchanged.
- **`view` is now effectively reserved.** A built-in shadows a *branch* declaration of the same name (as `event` and `command` already do), so an author who declares their own concept literally named `view` loses to the built-in in any later document. Existing spaces are unaffected: the built-in's `show` field is byte-identical to what the library declared, so their views resolve exactly as before and simply gain an optional field. Only a *differently shaped* author `view` breaks — which is what commit (5) fixes in the CLI's own fixtures, and what commit (1)'s fallout was in three `rule.rs` entity-derivation tests that used `view` as an incidental fixture name (now `card`). Note the asymmetry, which predates this PR: an *in-document* `concept!: &view` still shadows the built-in, so the reservation only bites across documents.

## Verification

CI's own invocation, which is `cargo nextest`, not `cargo test`:

```console
cargo fmt --all
cargo nextest run --workspace --exclude tonk-ui --exclude tonk-core \
    --features integration-tests --no-fail-fast \
    -E 'not test(it_suggests_sudo_when_the_target_directory_is_not_writable)'
# → 2005 tests run: 2005 passed, 4 skipped
```

Browser tests were run locally too, against a real Chrome — `tonk-display` 191/191 twice, `tonk-analyzer` 173/173, `tonk-template` 42/42.

The one filtered test is `tonk-cli`'s `it_suggests_sudo_when_the_target_directory_is_not_writable`, which cannot pass in a container running as root — nothing is unwritable. It passes on CI's unprivileged runner.

Two things worth flagging, because each is why a separate commit exists:

- `cargo test` stops launching further test **binaries** after one fails, so that root-only failure in `tonk-cli`'s lib tests silently prevented the `tonk-cli::site` integration binary from running at all. Every local `cargo test --workspace` was green only up to that point. `nextest --no-fail-fast` runs every binary regardless, and is what surfaced the three `site` failures behind commit (5).
- Commit (8) touches a crate this PR is otherwise unrelated to, so it deserves scrutiny. `tonk-fab`'s `set_parent_width` polled 50× at 10ms for a `ResizeObserver`-driven layout to settle, while `wait_for_width` ten lines above polls 500× at 10ms for the same kind of wait. The evidence that the short budget is marginal rather than the code broken: the identical `tonk-fab` binary — the crate links none of the code this branch changes beyond a static descriptor in `tonk-schema`, unchanged between these heads — passed `Tests (web, debug)` on `43e220c` at 1.875s and failed it on `9a508c8` at 2.378s, against 1.853s on the base. Same binary, three runs, both outcomes. Waiting longer costs nothing when the layout settles promptly: the loop returns on the first matching poll.

New tests:

- `tonk-template::bindings` — bare vs. quoted attribute values, a binding named inside an HTML comment, `xlink:`/`bind:` left alone, dag-cbor round-trip and byte-stability, a foreign `kind` refused rather than read as empty, and each binding's recorded offset landing on its attribute name (with a repeated binding keeping the first occurrence).
- `tonk-analyzer::view` — a binding is inlined with its `where:` source intact; a view binding nothing grows no artifact; each of the three diagnostics fires; a dangling binding's report covers the attribute alone rather than the block scalar; `as: record` is declarable.
- `tonk-analyzer::library_analysis_tests::it_inlines_the_bindings_the_shipped_libraries_make` — the pass is silent when it succeeds, so this asserts it still reaches real templates: every shipped library that binds an interaction must produce a decodable artifact, and `profile.yaml`'s create form must carry `on/space-create`.
- `tonk-display::element` — the artifact resolves with **no** query host installed (so it cannot have queried); a name it omits still goes to the branch; the stored artifact decodes back; foreign bytes resolve to nothing; and the state is not borrowed while the bindings query is in flight. The fixtures tear down their host element and their `body` listener on drop, since every wasm test in the crate shares one page and a stray `tonk-query` listener claims queries for whoever dispatches them.
- `tonk-worker::the_view_queries_match_the_builtin` — the display's hand-mirrored view predicate and the built-in cannot drift on `the` or cardinality (the parity `event` already had).

## Storybook impact

- [x] No user-visible contract changed because: this moves work from render time to lowering time. Every binding that worked before still works, resolved from a stored artifact instead of a per-name query; a view without the artifact takes the unchanged fallback path. What is new is that a *broken* binding in a template now fails the build instead of rendering an inert element — no working screen or journey changes. (The rename fix carried in from #904 restores a screen that was broken, and is described there.)

https://claude.ai/code/session_01WyWgReUaivze37mfWSNA4D

---

## PR-Codex overview
This PR introduces support for event bindings in the `tonk` framework, enabling views to declare and resolve events more effectively, improving the handling of user interactions.

### Detailed summary
- Added `bindings` module to handle event bindings.
- Introduced `EventBinding` struct for storing binding details.
- Updated `EventDescriptor` to include serialization support.
- Enhanced the `view` concept to include a `bindings` field.
- Modified the analyzer to index `event!:` declarations.
- Implemented logic to resolve bindings at lowering time.
- Updated tests to verify the functionality of inlined event declarations.

> The following files were skipped due to too many changes: `rust/tonk-analyzer/src/analyzer/view.rs`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4752,6 +4752,7 @@ dependencies = [
  "tonk-core",
  "tonk-notation",
  "tonk-schema",
+ "tonk-template",
  "wasm-bindgen-test",
 ]
 
@@ -5255,6 +5256,8 @@ dependencies = [
  "dialog-common",
  "indexmap",
  "ipld-core",
+ "serde",
+ "serde_ipld_dagcbor",
  "serde_ipld_dagjson",
  "serde_json",
  "tokio",
--- a/rust/tonk-analyzer/Cargo.toml
+++ b/rust/tonk-analyzer/Cargo.toml
@@ -33,6 +33,7 @@ thiserror = { workspace = true }
 tonk-core = { workspace = true }
 tonk-notation = { workspace = true }
 tonk-schema = { workspace = true }
+tonk-template = { workspace = true }
 
 [dev-dependencies]
 anyhow = { workspace = true }
--- a/rust/tonk-analyzer/src/analyzer.rs
+++ b/rust/tonk-analyzer/src/analyzer.rs
@@ -48,6 +48,7 @@ mod resolver_registry;
 mod rule;
 mod scan;
 mod scope;
+mod view;
 
 use std::collections::{HashMap, HashSet};
 
@@ -4931,6 +4932,75 @@ mod library_analysis_tests {
             "notebook.yaml on profile",
             &format!("{profile}\n{notebook}"),
         );
+    }
+
+    /// Every shipped library that binds an interaction must carry the
+    /// resolved bindings on its views.
+    ///
+    /// The inlining pass is silent when it succeeds, so
+    /// [`it_analyzes_the_shipped_libraries`] would pass just as well
+    /// if the pass stopped finding anything. This is the check that it
+    /// still reaches real templates: `on:` bindings exist in these
+    /// files, so `bindings` artifacts must too.
+    #[test]
+    fn it_inlines_the_bindings_the_shipped_libraries_make() {
+        let core = include_str!("../../tonk-core/assets/library/core.yaml");
+        for (name, body) in [
+            (
+                "notebook.yaml",
+                include_str!("../../tonk-core/assets/library/notebook.yaml"),
+            ),
+            (
+                "table.yaml",
+                include_str!("../../tonk-core/assets/library/table.yaml"),
+            ),
+            (
+                "prose.yaml",
+                include_str!("../../tonk-core/assets/library/prose.yaml"),
+            ),
+        ] {
+            let inlined = inlined_declarations(name, &format!("{core}\n{body}"));
+            assert!(
+                !inlined.is_empty(),
+                "{name} binds interactions, so its views must carry resolved bindings",
+            );
+        }
+        let profile = include_str!("../../tonk-core/assets/library/profile.yaml");
+        let inlined = inlined_declarations("profile.yaml", profile);
+        assert!(
+            inlined.contains(&"on/space-create".to_string()),
+            "profile.yaml's create form must carry its declaration inlined, got {inlined:?}",
+        );
+    }
+
+    /// Every `event!:` declaration name the document's views inlined.
+    fn inlined_declarations(name: &str, source: &str) -> Vec<String> {
+        use dialog_artifacts::Value as ArtifactValue;
+        use dialog_query::Term as QueryTerm;
+
+        let parsed = tonk_notation::parse(source);
+        let syntax = parsed
+            .syntax
+            .unwrap_or_else(|| panic!("{name} yields a syntax tree"));
+        let tree =
+            analyze_local(&syntax).unwrap_or_else(|error| panic!("{name} must analyze: {error:?}"));
+        let mut out = Vec::new();
+        for planned in tree.analysis.statements() {
+            let Statement::Assert(Application::Concept { query, .. }) = &planned.statement else {
+                continue;
+            };
+            let Some(QueryTerm::Constant(ArtifactValue::Bytes(bytes))) =
+                query.terms.get("bindings")
+            else {
+                continue;
+            };
+            let bindings = tonk_template::bindings::Bindings::decode(bytes)
+                .unwrap_or_else(|error| panic!("{name}'s artifact must decode: {error}"));
+            out.extend(bindings.events.into_keys());
+        }
+        out.sort();
+        out.dedup();
+        out
     }
 
     fn assert_analyzes(name: &str, source: &str) {
--- a/rust/tonk-analyzer/src/analyzer.rs
+++ b/rust/tonk-analyzer/src/analyzer.rs
@@ -4989,7 +4989,7 @@ mod library_analysis_tests {
             let Statement::Assert(Application::Concept { query, .. }) = &planned.statement else {
                 continue;
             };
-            let Some(QueryTerm::Constant(ArtifactValue::Bytes(bytes))) =
+            let Some(QueryTerm::Constant(ArtifactValue::Record(bytes))) =
                 query.terms.get("bindings")
             else {
                 continue;
--- a/rust/tonk-analyzer/src/analyzer/assertion.rs
+++ b/rust/tonk-analyzer/src/analyzer/assertion.rs
@@ -336,6 +336,34 @@ pub(crate) fn build_assertion_application(
                 }
             }
 
+            // A view's `on:<name>=<command>` bindings are resolved
+            // here, not at render time: the declaration and the
+            // command are both in scope, so an unresolvable reference
+            // fails the lowering instead of producing an inert
+            // element. The resolved table rides the view as one CBOR
+            // artifact, so the display decodes it rather than issuing
+            // a query per declaration name on every refresh.
+            if super::view::is_view(&resolved) {
+                user_fields.remove(super::view::BINDINGS_FIELD);
+                if let Some(bindings) = super::view::compile_bindings(assertion, scope)? {
+                    let encoded = bindings.encode().map_err(|reason| {
+                        AnalyzeError::at(
+                            AnalyzeErrorKind::InvalidViewBindings { reason },
+                            head_range,
+                        )
+                    })?;
+                    assert_terms.insert(
+                        super::view::BINDINGS_FIELD.into(),
+                        Term::Constant(Value::Bytes(encoded)),
+                    );
+                    retract_terms.insert(
+                        super::view::BINDINGS_FIELD.into(),
+                        Term::<dialog_query::Any>::blank(),
+                    );
+                    any_assert = true;
+                }
+            }
+
             if let Some((unknown, _)) = user_fields.into_iter().next() {
                 let unknown_range = assertion
                     .fields
--- a/rust/tonk-analyzer/src/analyzer/assertion.rs
+++ b/rust/tonk-analyzer/src/analyzer/assertion.rs
@@ -354,7 +354,7 @@ pub(crate) fn build_assertion_application(
                     })?;
                     assert_terms.insert(
                         super::view::BINDINGS_FIELD.into(),
-                        Term::Constant(Value::Bytes(encoded)),
+                        Term::Constant(Value::Record(encoded)),
                     );
                     retract_terms.insert(
                         super::view::BINDINGS_FIELD.into(),
--- a/rust/tonk-analyzer/src/analyzer/declaration.rs
+++ b/rust/tonk-analyzer/src/analyzer/declaration.rs
@@ -937,6 +937,11 @@ fn normalize_type_name(name: &str) -> Option<&'static str> {
         "boolean" | "Boolean" => Some("Boolean"),
         "entity" | "Entity" => Some("Entity"),
         "bytes" | "Bytes" => Some("Bytes"),
+        // Dialog's type for "one struct, one fact" — a value whose
+        // bytes are a serialized record rather than an opaque blob.
+        // Spellable here because the built-in `view` declares one and
+        // an author's concept must be able to say the same thing.
+        "record" | "Record" => Some("Record"),
         _ => None,
     }
 }
--- a/rust/tonk-analyzer/src/analyzer/error.rs
+++ b/rust/tonk-analyzer/src/analyzer/error.rs
@@ -350,6 +350,51 @@ pub enum AnalyzeErrorKind {
         /// The concept name that failed to resolve.
         name: String,
     },
+    /// A `on:<name>=<command>` binding in a view template names an
+    /// `event!:` declaration nothing declares. The binding would
+    /// install no listener and report nothing — the element would
+    /// simply be inert — so it fails the lowering instead.
+    #[error(
+        "`{attribute}` names the event declaration `{name}`, which no `event!:` declares          in this document or on the branch"
+    )]
+    UnknownEventDeclaration {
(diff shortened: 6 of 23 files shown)


Gozala:
**Blocked on a re-run: `Tests (web, debug)` failed on `5190fbd`, and I judge it not this PR's.**

The reasoning, so it can be checked rather than taken on trust:

- The web archive is built `--workspace --exclude tonk-cli` (`flake.nix`, `tests-web-debug`).
- The diff from `43e220c` (the previous head) to `5190fbd` is **only** `rust/tonk-cli/tests/common.rs` and `rust/tonk-cli/tests/site.rs` — two `tonk-cli` integration-test files, in the one package that archive excludes.
- `Tests (web, debug)` **passed** on `43e220c` an hour earlier. Same wasm test binaries, opposite result.

The job's cleanup also had to terminate orphaned `wbg-pool`, `chromium` and two `chrome_crashpad_handler` processes, which points at the browser pool rather than at a test body. The wasm-running tests this PR adds (`tonk-template::bindings::*`, `tonk-analyzer::view::*`) are pure computation — a string scan, a dag-cbor round-trip, `analyze_local` — with no timers, IO, or browser APIs, so they have nothing to flake on.

**What I need:** a re-run. `rerun-failed-jobs` on run `34008395792` returns `403 Resource not accessible by integration` for me, and I will not push an empty commit to kick CI. Someone with write access re-running that run unblocks it.

That re-run also settles something this run did not: `Tests (native, debug)` — the job that caught the real failure on the previous head, and the one `5190fbd` exists to fix — was **cancelled** by fail-fast at ~17 minutes, so it never reported. Locally, CI's own invocation is green:

```console
cargo nextest run --workspace --exclude tonk-ui --exclude tonk-core \
    --features integration-tests --no-fail-fast \
    -E 'not test(it_suggests_sudo_when_the_target_directory_is_not_writable)'
# → 2003 tests run: 2003 passed, 4 skipped
```

(the filtered test cannot pass in a container running as root — nothing is unwritable — and passes on CI's unprivileged runner).

I am attempting to reproduce the wasm run locally to rule myself out properly rather than resting on the diff argument. I will report either way and keep watching this PR until it is green.

---

Gozala:
Second browser job down on the same head: **`End-to-end shard 2/3`** (shards 1 and 3 passed; "End-to-end account tests" then failed in 2s as its dependent). Same conclusion as the `Tests (web, debug)` note above, and now with better evidence than the diff argument.

**The whole feature was already green.** `43e220c` — the previous head — contained all four substantive commits (built-in `view`, the library change, the inlining pass, `Record`). On that head **all three e2e shards passed**, at 02:19–02:25. The only change since is `rust/tonk-cli/tests/{common,site}.rs`, two integration-test sources that compile into neither the `tonk` binary (`nix build .#tonk-cli`) nor the e2e archive (`--package tonk-ui`). The e2e inputs are the same ones that passed.

From the `e2e-diagnostics-2` artifact, shard 2's 17 tests are 15 `account_flow::*` and 2 `service_worker_upgrade::*`, and the last to start — inferred from pid ordering, since the phase logs carry no failure marker — was `service_worker_upgrade::tests::it_reloads_every_update_aware_tab_after_controller_replacement`. Nothing in that shard renders a template or exercises an `on:` binding.

This is also the mode `.config/nextest.toml` already documents:

> a shared CI runner is 2-3x slower than the hand-rolled deadlines in these tests assume: the same commit has passed every shard in one run and failed three in another, tripping a different deadline each time

Both failing jobs in this run also left orphaned `chromium` / `chrome_crashpad_handler` processes for the runner to reap — five browsers' worth in the e2e shard.

**Following up on the last comment:** I tried to reproduce the wasm suite locally rather than rest on the argument, and could not — this container's chromedriver is `SIGKILL`ed at startup (no IPv6; `CreatePlatformSocket() failed: Address family not supported by protocol`). Version-matching the driver to Chromium 141 did not help. That is my sandbox, not the code, and I stopped there rather than keep digging.

**Still the same ask:** a re-run of run `34008395792`. `rerun-failed-jobs` gives me `403 Resource not accessible by integration`, and I will not push an empty commit to kick CI. That re-run also settles `Tests (native, debug)`, which `5190fbd` exists to fix and which fail-fast cancelled before it could report.

---

Gozala:
**Retracting the two comments above — those failures were this PR's, and `9a508c8` fixes them.**

`refresh_delegate` had:

```rust
let inlined = match state.borrow().model_entity.clone() {
    Some(model_entity) => resolve_inlined_bindings(host, &model_entity).await,
    None => BTreeMap::new(),
};
```

A `match` keeps its scrutinee's temporaries alive for the whole match, so that `Ref` was held across the awaited query I added. Anything that fires during that window — a view frame, an entity frame, a disconnect — calls `state.borrow_mut()` and panics. Snapshotting the entity in its own statement drops the `Ref` at the end of that statement, which is the shape the view-element snapshot a few lines above already uses for exactly this reason.

Where my reasoning went wrong is worth naming, since it was the same argument twice: I concluded "not this PR's" from *the artifacts are identical to `43e220c`, where these jobs passed*. That is not evidence of anything when the defect is a race — a race passes sometimes, which is why it passed there and why two browser jobs failed here while the other two passed. Staging being green was the tell I should have taken seriously instead of reaching for the repo's flake note.

Verification on `9a508c8`:

```console
cargo nextest run --workspace --exclude tonk-ui --exclude tonk-core \
    --features integration-tests --no-fail-fast \
    -E 'not test(it_suggests_sudo_when_the_target_directory_is_not_writable)'
# → 2003 tests run: 2003 passed, 4 skipped
cargo test -p tonk-analyzer -p tonk-template -p tonk-display -p tonk-worker \
    --no-run --target wasm32-unknown-unknown
```

Neither pins the fix — it is a browser-timing bug and the display's refresh path has no test harness — so CI on this head is the real check. The unrelated re-run request in those comments is moot: this push starts a fresh run.

---

PR #918, opened 2026-09-09
fix(worker): use disposable bounded in-memory sessions

jackddouglas:
## What changed

Fix worker-session authorization after browser worker replacement. Safari can produce different signatures for identical inputs, which changed the reconstructed operator DID while the worker reused a persisted grant addressed to its previous operator. Each boot and renewal now creates a fresh random operator with a matching 12-hour grant held in memory.

The change preserves account/profile membership, invitation authority, local contents, the one-hour renewal margin, and existing legacy session records. Renewal installs operator and expiry atomically; failed or losing concurrent candidates create no durable session writes. Session construction no longer needs the old delegation-hydration fallback.

Depends on https://github.com/dialog-db/dialog-db/pull/489. The Dialog family is pinned to tested commit `314e49926eb5bca58bacdcf9f2123ccb4422ea35`, with the corresponding Nix hash; the separate runner input is unchanged.

## Verification

All checks below used the published immutable Dialog pin without local overrides:

- `cargo test -p tonk-worker --lib --locked`: 121 passed.
- Chromium `wbg-pool`: 11 tests matching `session::tests`, three `router::sync::renewal_tests`, and 26 `router::join::tests` passed.
- `nix build .#checks.aarch64-darwin.sharedWorkspaceDeps --no-link`: passed.
- Rust/Nix formatting, Nix source-reference, and diff whitespace checks passed.

Regression coverage includes distinct operators across opens, bounded proofs without profile commits, storage close/reopen with mismatched legacy authority, forced renewal, construction failure, concurrent candidates, and preserved onboarding membership after rebuilding worker state.

macOS Safari/iOS Safari, actual remote pull/push after worker replacement, network-offline restart, and signing/boot/pull latency measurements remain unverified. The onboarding fixture has no remote; its passing proof/content assertions do not establish network replication. This change has not been deployed.

## Storybook impact

User-visible sync reliability fix with backend regression coverage in `session::tests`, `router::sync::renewal_tests`, and `it_reopens_an_onboarding_membership_with_a_disposable_session`. No rendered UI or CLI output changed. Storybook was not updated and no shared Storybook regression ID was added; device acceptance remains recorded in `plan/in-memory-worker-sessions.md`.

---

## PR-Codex overview
This PR focuses on creating disposable worker operators with bounded in-memory grants, improving session management in the `tonk-worker` by ensuring that sessions are created fresh on each boot without relying on persisted data, thus enhancing reliability and performance.

### Detailed summary
- Updated `cargoGitDependencies` to use `rev` instead of `tag` for consistency.
- Refactored session handling in `rust/tonk-worker/src/worker.rs` to avoid reliance on persisted sessions.
- Introduced a new test in `rust/tonk-worker/src/router/join.rs` for onboarding membership with disposable sessions.
- Enhanced session renewal logic in `rust/tonk-worker/src/router/sync.rs` to create disposable operators.
- Removed old session persistence code and related structures in `rust/tonk-worker/src/session.rs`.
- Added tests for distinct session creation and proof validation in `rust/tonk-worker/src/session.rs`. 
- Documented the new session management strategy in `plan/in-memory-worker-sessions.md`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -986,7 +986,7 @@ dependencies = [
 [[package]]
 name = "dialog-artifacts"
 version = "0.1.0"
-source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-09-02#805151c678c516edc03c2825171ed4b54adf37ab"
+source = "git+https://github.com/dialog-db/dialog-db.git?rev=314e49926eb5bca58bacdcf9f2123ccb4422ea35#314e49926eb5bca58bacdcf9f2123ccb4422ea35"
 dependencies = [
  "arrayref",
  "async-stream",
@@ -1025,7 +1025,7 @@ dependencies = [
 [[package]]
 name = "dialog-capability"
 version = "0.1.0"
-source = "git+https://github.com/dialog-db/dialog-db.git?tag=tonk-2026-09-02#805151c678c516edc03c2825171ed4b54adf37ab"
+source = "git+https://github.com/dialog-db/dialog-db.git?rev=314e49926eb5bca58bacdcf9f2123ccb4422ea35#314e49926eb5bca58bacdcf9f2123ccb4422ea35"
 dependencies = [
  "async-trait",
  "base58",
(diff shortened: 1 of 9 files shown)


Gozala:
> Safari can produce different signatures for identical inputs, which changed the reconstructed operator DID while the worker reused a persisted grant addressed to its previous operator

I think we need to fix this issue, otherwise we inflate number of origins and dialog will need to continue to track each one in the watermark

PR #937, opened 2026-09-15
fix(tree): history, coverage and blob rows read as what they are in dialog:diagnose

Gozala:
## What changed

The `dialog:diagnose` tree visualizer showed history records as blank rows — no entity, no value, with the word `history` squatting in the Attribute column — and rendered blob-index nodes as garbage. Both regions now read as what they are.

- **History and coverage records carry their fact.** `tree/entry` reconstructed a row only from EAV-shaped keys. A history or coverage key is the same fact behind a fixed-width version prefix (`tag ‖ origin(32) ‖ edition(8) ‖ entity ‖ attribute ‖ value slot`), so dropping the prefix and re-tagging the tail reconstructs it — the same move dialog's own `key_components` makes internally.
- **A history row names the revision that wrote it.** The row's version chip reads `<origin head>@<edition>`, with the whole origin in its tooltip and split out in the detail pane. The edition alone is a causal depth, not an identity: two records at the same edition under different origins are concurrent writes by different replicas, and showing the edition by itself made those rows read as duplicates.
- **Blob rows name their content.** The hash now travels as the `#<base58>` the rest of `tree/*` uses and as the `blob:<base58>` entity the same bytes carry their `xyz.tonk.blob/*` facts under, instead of a raw hex run that cross-referenced nothing. An undecodable blob record no longer propagates its error and blanks the whole segment.
- **Separators decode.** A blob separator was passed through as utf8-lossy raw hash bytes — a run of replacement characters across the outline; it now reads as a hash prefix. The fact behind a history separator splits into its own coloured fields instead of one long entity.
- **"0x05 index" is gone.** The index and value-type chip tooltips were built from the part's raw `hex` byte rather than its decoded `text`, so a coverage key's chip explained itself as `0x05 index`. They now read `Coverage index`, `Value type: String`, and so on.
- **A spilled value stops lying about its type.** The placeholder artifact stands a `<spilled value>` *string* in for bytes it cannot fetch; the row now reports the type the key records and names the block the value lives in.
- **Novelty is visible on nodes.** The outline has a novelty column: `⊕n` for what a node buffers for its own children, `↓n` for what its parent still buffers against that span (carried on `tree/child` rows as `pending`, so it is known even for an unfetched child). The inspector labels both. Related: expanding a node re-read it through `tree/node` and overwrote the row, dropping its decoded separator back to a bare hash fragment — span-derived fields are now carried across that refresh.
- The entries table gains an **Index** column, which is what frees Entity / Attribute / Value to say what the row actually holds.

## Verification

```console
cargo test -p dialog-reactor -p tonk-tree --lib     # 54 pass
cargo clippy --all --all-targets --all-features -- -D warnings
cargo fmt --all
cargo check -p tonk-tree --target wasm32-unknown-unknown
```

New coverage: `it_decodes_every_region_of_a_mixed_segment` builds a real segment node holding a fact, a history record, a coverage record and a blob reference, and runs the whole `tree/entry` decode over it; plus per-case tests for the history/coverage reconstruction, blob naming and hex fallback, separator decoding, and the chip labelling in `tonk-tree`.

Not run here: the worker's router tests are wasm-only and this environment has no `wbg-pool`/browser runner, so the endpoints were not exercised through a seeded branch. The decode was pushed down to a pure `segment_rows(bytes)` and tested against real node bytes instead. The rendering itself (column widths, chip layout) is unverified visually.

## Storybook impact

- [x] User-visible browser or CLI behavior changed. I updated these screen, journey, verification, or triage IDs: **none — no Storybook ID covers the `dialog:diagnose` tree inspector today** (`screens.json` has no entry for it, and `docs/storybook` mentions `diagnose` only in the `/inspector` summary paragraph). The regression coverage lives with the code, in `dialog-reactor`'s `formula` tests and `tonk-tree`'s `key` tests. Say the word if this surface should get an ID and I will add one.

## Known divergence: this is not the resolver path

Worth stating plainly, because the names collide. tonk already uses dialog's tree **resolvers** — but on the notation/eval path only (`tonk-analyzer`'s resolver registry, `QueryPlan::Resolver`, `Application::Resolver`, and `dialog/key-part` / `dialog/separator-part` as registered formulas). `<tonk-tree>` is not on that path: it POSTs `{"predicate": "tree/child", …}` to `/query`, and `router/query.rs` has exactly two arms — a concept predicate goes to dialog's planner, a string predicate goes to `dialog-reactor::resolve_formula`. That worker-side `tree/*` family decodes with `inspect_*` directly and hand-rolls its own `read_node`/`NetworkedIndex`, bypassing the shared node cache dialog's `QueryEnv: Provider<Load>` reads through.

Same names, different shapes:

| tonk formula (`dialog-reactor/src/formula.rs`) | dialog resolver |
| --- | --- |
| `tree/node` (+ manifest, bound, rank inline) | `tree/node` + `tree/manifest` |
| `tree/child` | `tree/span` (joined into `tree/node`) |
| `tree/entry` (key + claim + blob + fact reconstruction in one row) | `tree/key` + `tree/entry` + `tree/blob` + `dialog/key-part` |
| `tree/key` | `dialog/key-part`, and `dialog/separator-part` for boundaries |

Two notes for whoever migrates it:

- The history blank-row bug this PR fixes is an artifact of the custom path specifically — it comes from its `Artifact::from_key_datum_placeholder` step, which rejects any tag but EAV/AEV/VAE. The resolver model has no such step: the fact comes out of `dialog/key-part`, which already decodes history, coverage and blob keys correctly. The chip-labelling and separator-splitting fixes here are client-side presentation and carry over either way.
- `cached` (the outline's filled/hollow locality dot) has **no** resolver equivalent, by design: a locality probe is not idempotent, so `dialog_artifacts::inspect` explicitly refuses to serve it through `Load`. A migration has to either give it its own effect or drop the dot.

https://claude.ai/code/session_01UeRL9ntUK6xpZS2MYCaBXX

---

## PR-Codex overview
This PR introduces new dependencies, enhances the `TreeNode` structure with a `pending` field, updates the UI to display novelty information, and refines the handling of entries in the segment tree, particularly for blob records and history entries.

### Detailed summary
- Added `dialog-common`, `tokio`, and `wasm-bindgen-test` as dependencies.
- Introduced `pending` field in `TreeNode` structure.
- Updated `carry_span_from` method to include `pending`.
- Enhanced UI to show novelty in rows.
- Improved entry handling for blob records and history entries.
- Added tests for index and value-type chips.
- Updated styles for novelty and index columns in the UI.


Gozala:
CI is red on **End-to-end shard 1/3**, and it is not this PR's failure. ("End-to-end account tests" is not a second failure — its whole body is `test "$SHARD_RESULT" = success`, so it mirrors the shard.)

The failing test is `tonk-ui account_flow::tests::it_deletes_the_account_and_releases_its_email_and_profile`, three retries, same assertion each time:

```
Error: deletion consent={"awaitingChoice":false,"message":null,"present":false}
Caused by: the page never landed on /; it is at https://tonk.network:40455/settings
```

That is the passkey bug #941 root-causes. The test drives `driver_with_prf` and fails in `custody_consent_diagnostic` — a custody assertion — and #941 describes exactly this symptom: the allow-list id is a view of freed wasm memory, the virtual authenticator answers CTAP `0x2E`, "the page could only report a dismissed prompt", and "locally about one custody assertion in five failed (account deletion, and the custody consent after a space create), which is exactly what the e2e retries had been hiding". With that fix it reports 20 of 20 local passes.

Two things corroborate it being unrelated to this diff:

- This branch's run at `02f73b67` (the same two commits over an earlier staging merge) passed shard 1. Same code, different heap layout.
- Every other job here is green, and the diff is the tree inspector — `dialog-reactor`'s `tree/*` formulas and `<tonk-tree>`. Nothing in the account, custody or settings path.

I have not ported #941 into this branch. It is 206 lines across WebAuthn, the e2e harness and `flake.nix`, authored here and already open against `staging`, so carrying it would make this PR unreviewable and conflict when #941 lands. The cheaper sequence is to land #941 and merge `staging` again.

Spending one re-run now, since the bug is heap-layout dependent and a green shard is likely; if it comes back red with the same assertion, that changes nothing about the diagnosis but I will say so rather than re-running again.

---

Gozala:
Correction to the previous comment: I cannot spend that re-run — `rerun-failed-jobs` returns 403 for this integration. The diagnosis is unchanged, and the failing test is not one this PR can make robust (the bug is in `tonk-identity`'s WebAuthn buffers, which #941 fixes). So this stays as it is until #941 lands and `staging` is merged again, or until someone with re-run rights kicks the shard. I will keep checking back rather than leaving it.

---

PR #944, opened 2026-09-15
fix(display): forward routing context into view template content

### LINKED ISSUE #942: Routing context is lost for consumers inside view templates (`no repository in context`)
## Symptom

Space-scoped inspector and tree routes render an error instead of their content:

- `/space/{did}/inspector` → `no repository in context (nest under a with="branch@repo" element)`
- `/space/{did}/dialog:diagnose` → same, via `<tonk-tree>`

Reproduced on staging and locally.

## What actually breaks

`<tonk-display>` resolves the correct context, but never delivers it to the consumer. Observed in the sealed guest for `/space/{did}/inspector`:

```
tonk-display  model="inspector"  data-state="empty"
              with="main@did:key:z6Mktf…"        ← display HAS the context
  └─ tonk-view      with=null
     └─ tonk-inspector  with=null                 ← consumer does NOT
```

## Cause

`forward_with` (`rust/tonk-display/src/element.rs`) propagates context by querying the mounted view:

```rust
view_el.query_selector_all("[data-tonk-with], tonk-display, tonk-tree, tonk-inspector, tonk-notebook, ui-sync-status")
```

`8acaa189` ("feat: parallelize block replication", #924) changed `mount_view_slide` so a view's markup is parsed into an **inert `<template>`** instead of into the view's live children:

```rust
let template = document.create_element("template").ok()?;
template.set_inner_html(display);
let _ = view_el.append_child(&template);
forward_with(host, &view_el);        // ← next line
```

That change is correct on its own terms — it stops custom elements connecting as the template instance, which was a real bug. But a `<template>`'s nodes live in its `content` DocumentFragment, and `querySelectorAll` **does not descend into it**. So `forward_with` now matches zero elements and stamps nothing, silently: no error, no warning.

The renderer later `clone_node_with_deep`s that template, so every clone inherits the missing attribute.

## Blast radius

Any routing consumer authored inside a view template, not just the two routes above: `<tonk-tree>`, `<tonk-inspector>`, `<tonk-notebook>`, `<ui-sync-status>`, and nested `<tonk-display>` — the last of which is how context is meant to cascade, so multi-level views lose it too.

Consumers that independently observe `with` (`<tonk-notebook>`, `<ui-sync-status>`) recover when the attribute lands. `<tonk-inspector>` (`observed_attributes` = `&[]`) and `<tonk-tree>` (`&["repo", "branch"]`) do not, so they latch the error permanently.

## Bisect

Last known-good reported: `faaf46df` (#933). `8acaa189` is its direct child and the only commit in the `faaf46df..staging` window that touches the mount path.

One caveat worth recording: it was also reported as working at `8acaa189` and broken at `824de497`. That is hard to reconcile with the code — `824de497` is a 6-line change to `tonk-account/src/delegations.rs` adding `.operational()` to an account pull, touching nothing on the routing or view-mount path, and `8acaa189` is its ancestor. A cached space or a stale served bundle would explain the discrepancy, but it has not been confirmed either way.

## Fix

Stamp the template content as well as the light-DOM subtree, before the renderer clones it. Secondarily, have `<tonk-inspector>` and `<tonk-tree>` observe `with` and mount when context lands, matching what `<tonk-notebook>` and `<ui-sync-status>` already do — so a late-arriving context is recoverable rather than latched.

Separately: `/space/{did}/diagnose` (no `dialog:` prefix) is not a valid space path — the space branch has no `/diagnose` route and no `diagnose` model, only `dialog:diagnose`. It resolves `model="diagnose"`, matches nothing, and renders empty. Working `/diagnose` on the profile branch is likely why it is expected to work on a space too.

Gozala:
Fixes #942.

## Why

`/space/{id}/inspector` and `/space/{id}/dialog:diagnose` render `no repository in context` instead of their content.

`<tonk-display>` resolves the right context but never delivers it. In the sealed guest:

```
tonk-display  model="inspector"  with="main@did:key:z6Mktf…"   ← display HAS context
  └─ tonk-view        with=null
     └─ tonk-inspector  with=null                              ← consumer does NOT
```

`forward_with` propagates context with `view_el.query_selector_all(…)`. #924 changed `mount_view_slide` to parse a view's markup into an inert `<template>` rather than into the view's live children — correct on its own terms, it stopped custom elements connecting as the template instance. But a template's nodes live in its `content` DocumentFragment, which `querySelectorAll` does not descend into. So the forward matched nothing and stamped nothing, silently, and the renderer then cloned the context-less template.

This affects any routing consumer authored in a view template, not just those two routes: `<tonk-tree>`, `<tonk-inspector>`, `<tonk-notebook>`, `<ui-sync-status>`, and nested `<tonk-display>` — the last being how context is meant to cascade, so multi-level views lose it too.

## How

Stamp the template content as well as the light DOM, before the renderer clones it. The two passes query from different roots (`Element` vs `DocumentFragment`) and are otherwise identical, so they share one helper.

Secondly, `<tonk-inspector>` (`observed_attributes` was `&[]`) and `<tonk-tree>` (`&["repo", "branch"]`) now observe `with` and mount when it lands. `<tonk-notebook>` and `<ui-sync-status>` already take that cue — its absence is why these two latched the error permanently instead of recovering. That makes a late-arriving context recoverable regardless of stamping order.

## Verification

Against a live dev server, driven over CDP, with the served guest bundle confirmed to contain the change:

| Route | Before | After |
|---|---|---|
| `/space/{id}/inspector` | `no repository in context` | `<tonk-inspector with="main@did:key:z6Mktf…">`, renders branch diagnostics |
| `/space/{id}/dialog:diagnose` | `no repository in context` | `<tonk-tree with="main@did:key:z6Mktf…">`, shadow root populated |
| `/inspector` (profile) | worked | still works |

No error sections in any frame. A regression test covers the template-content path at the `forward_with` level, so this cannot silently recur.

## Note

`/space/{id}/diagnose` (without the `dialog:` prefix) is a separate matter and still renders empty: the space branch has no `/diagnose` route and no `diagnose` model, only `dialog:diagnose`. `/diagnose` working on the profile branch is likely why it is expected to work on a space too.

---

## PR-Codex overview
This PR focuses on enhancing the context propagation for routing consumers in the `tonk` framework by ensuring that the `with` attribute is properly observed and stamped onto elements within templates, preventing context loss during rendering.

### Detailed summary
- Modified `observed_attributes` in `rust/tonk-tree/src/web.rs` to include `with`.
- Updated `observed_attributes` in `rust/tonk-inspector/src/element.rs` to observe `with`.
- Added `attribute_changed_callback` in `rust/tonk-inspector/src/element.rs` to handle context changes.
- Enhanced context stamping logic in `rust/tonk-display/src/element.rs` to include template content.
- Introduced a test to verify that `with` is correctly forwarded into template content.


Gozala:
Closing as a duplicate of #943, which landed first and is the better implementation: it recurses into nested templates (a view's row template nests inside the mount's), which this PR's top-level-only pass would miss.

Same diagnosis, independently reached — see #942 for the write-up and the runtime verification (`/space/{id}/inspector` and `/space/{id}/dialog:diagnose` both confirmed fixed against a live dev server with the served guest bundle checked for provenance).

One piece here is not in #943 and may be worth carrying over separately: `<tonk-inspector>` (`observed_attributes` was `&[]`) and `<tonk-tree>` (`&["repo", "branch"]`) do not observe `with`, so they latch the error permanently instead of recovering when context lands late. `<tonk-notebook>` and `<ui-sync-status>` already observe it — that asymmetry is why only those two elements stayed broken on screen. Independent of the stamping fix, but it makes late context recoverable rather than fatal.

PR #970, opened 2026-09-19
feat(hub): the hub bar is branch data, and ui-hub-account is gone

Gozala:
## Why

`<ui-hub-account>` was 2,402 lines of Rust drawing the hub's account bar: the name, the profile switcher, tab switching, menu keyboard navigation, and the signup ceremony's position. Every part is now a view over facts or an `element!:` definition the branch carries, and the file is deleted. The settings panel follows the same road and loses ~290 lines.

## How it splits

**Facts** — the account name is `tonk:account/name`, the switcher list is `tonk:profile/row`, the settings pane renders `tonk:account/registered`, and each passkey is a `tonk:account/passkey` row.

"A signup is up" is a `CeremonyStatus` row, and that one is the interesting part. The hub held the flag on the **document body**, with this comment explaining why:

> The route view re-renders whenever profile facts land (the passkey itself lands one), replacing this element mid-ceremony.

`CeremonyStatus` already existed for exactly this, written by `ceremony::report` and read by the settings panel. The linking flow never adopted it. An overlay row survives a re-render by not living in the DOM at all, so the workaround goes rather than being reimplemented.

**Behaviour** — `<hub-bar>` owns tab state and the ceremony seat; `<hub-menu>` owns roving focus, Escape, and returning focus to its opener. Both are `element!:` definitions, so repointing a tag swaps the behaviour for every instance already on the page.

The bar must measure its own rectangle: the hub is a sealed guest at an opaque origin, so the top page that draws the ceremony cannot reach in.

**Commands** — switching and adding a profile were `POST /api/profiles/*`; they are `tonk:switch-profile` and `tonk:add-profile`, landing in the same handlers the routes use. The display-name save was a POST to an endpoint that calls the same `rename_display_name` as `profile/rename`, so it is that command now.

## One duplication, on purpose

The `element` concept is repeated in `profile.yaml` rather than referenced from `core.yaml`. The hub is sealed to the profile meta branch and resolves a definition by querying **that** branch, the same reason `profile/rename` is duplicated there.

## A test gap, closed

`enter_ends_a_display_name_edit` tested the element's own keydown handler. The gesture moved to `<tonk-editable>`, which had **no tests at all**, so it gains two: Enter ends the edit so the change-save path runs, and Escape restores what the edit started from. Both begin the edit by double-click — focusing alone tests nothing, since `contenteditable` starts false so a stray click cannot open an edit. The Enter test is mutation-verified: removing the blur fails it.

## Verified on the live page

Both hub routes, after the deletion: both tags register from the branch, `customElements.get('ui-hub-account')` is undefined, clicking the account cell switches the tab and moves `aria-current`, the menu opens onto its first focusable row and moves through them, `seat()` returns a real rectangle, and `/settings` mounts the view with the old input gone.

2212 native, 1778 wasm, clippy clean.

## What still needs an element

Three passkey ceremonies and the deletion flow: each makes a live `navigator.credentials` call no view can.

---

## PR-Codex overview
This PR focuses on refactoring the UI components of a Rust-based application, specifically replacing the `ui-hub-account` element with a new `hub-bar` element, which manages account-related functionalities. It also introduces a new profile handling system and updates the related HTML structure.

### Detailed summary
- Deleted files: `hub_account.rs`, `ui_hub_account.rs`, `ui_hub_account.html`.
- Introduced `hub-bar` element to manage account tab and linking states.
- Added `ADD_PROFILE` constant to `account.rs` for profile signup.
- Updated `profiles.rs` to report signup status to the hub.
- Refactored `ui_account_settings` to use a `tonk-display` for account facts.
- Removed passkey management from `ui_account_settings.rs`.
- Adjusted tests to reflect changes in the new structure and behavior.


Gozala:
Folded into #986: these three only go green together with the elements rewrite and the settings port, so they merge as one PR on top of #969. The commits are unchanged and all ride on #986.

PR #990, opened 2026-09-22
refactor: port to dialog-peer, pinned to the Peer/Session branch

Gozala:
## What changed

dialog-db/dialog-db#527 renames `dialog-operator` to `dialog-peer`, splits `Operator` into a stored `Peer` and a scoped `Session`, and drops `Profile`. This ports tonk to that API with every dialog dependency pinned to that branch. Once #527 merges and is tagged, the pins move to the tag; nothing else in this PR needs to change for that.

The peer is the identity:

- Every `Profile` tonk held is now the `Peer` it was opened from: `tonk_cli::peer::NativePeer` in the CLI, `DefaultPeer` in the worker. Peers open through the builder at a `Location` (`open`, `load`, `create`), and the reactor holds the signer credential rather than a profile.
- Sessions derive with `peer.derive(ctx).await?.allow(..).build()`. Where the CLI rebased an operator at a data directory, `tonk_cli::peer::peer_for` builds a peer over the same credential and storage with that base, and `session_for` derives from it.
- `profile.signer()` is `peer.credential()`, `profile.repository(..)` is `peer.space(..)`, `profile.credential().site(..)` is `peer.secrets().site(..)`, `operator.profile_did()` is `session.peer().did()`: the replica identity, not the acting key.
- The worker rotates sessions over its peer with an expiring claim, in place of `allow_until`, and its renewal builder takes the peer alone.
- The site no longer persists a fresh signing grant on every open. The grant is minted in memory at build, which is what the in-memory session was introduced for.
- The invite key is generated as `Ed25519Signer<Extractable>` on every platform, since dialog-db#520 made that its own type and the wasm arm asked the sealed type for it.

The scoped-invitation series (#976 to #989) landed mid-port and is merged in, with its derivation sites ported the same way.

No user-visible behaviour changes. Session keys derive under the same label #520 pinned, so a session derived from a context is the operator that context derived before.

## Verification

On the head with the pins at dialog-db `b7c77af`, from a clean build directory:

- `cargo clippy --workspace --all-targets --all-features -- -D warnings` passes.
- `cargo fmt --all --check` passes.
- `cargo check --workspace --exclude tonk-cli --exclude tonk-perf --all-targets --target wasm32-unknown-unknown` passes.
- `cargo test --workspace --exclude tonk-ui --exclude tonk-core --features integration-tests`: 2324 pass, 1 fails. The failure is `update::swap::tests::it_suggests_sudo_when_the_target_directory_is_not_writable`, which fails because the container runs as root and ignores the directory mode. That file is untouched.
- Not run locally: the e2e suite and the browser wasm tests, which need nix. CI on this head runs them and every job is green, including the service-worker lifecycle job whose one failing Node test on an earlier head was unrelated to this diff.

## Storybook impact

- [x] No user-visible contract changed because: this is a dependency API migration with no change to CLI output, browser flows, or stored data.

https://claude.ai/code/session_01SyWN8CCDeBinXiRVt8pBYo


Gozala:
The **Service-worker lifecycle tests** check failed on `c8a8760` in one Node test, `the checked-in tonk-code bundle holds reconnect while an update is pending` (`rust/tonk-ui/tests/build-artifacts.test.mjs:321`): it saw one reconnect after the controller change where it expects two.

That is not this PR's. The diff touches no JavaScript, no asset, and no test file; the bundle and the test it runs were last changed in #931 and are byte-identical to staging, where the same job is green on the latest push. The test also fails the same way locally on those identical inputs, inside a fixed two-second window, so it is sensitive to the machine it runs on rather than to this branch.

No fix is ported, since nothing in this PR can move it. The next push re-runs the job.

**Resolved on `f252afa`:** the job passed on the re-run, along with every other check on that head.

---

Gozala:
The red checks on `a1bbce2`, `f7e277e` and `3da1ada` are expected. Those commits port tonk to the next dialog-db change on the same branch, dialog-db#527 dropping `Profile`, while the pins still name the revision before it, so the lint, CLI build and e2e jobs cannot compile. The pin update lands with the next push once the dialog-db side is verified, and CI is meaningful again from that head.

**Resolved on `f252afa`:** the pins now name dialog-db `b7c77af`, and every check on that head is green.

---

Gozala:
dialog-db#527 has since merged `main`, which brought in dialog-db#524 (the capability method rename). This PR stays pinned at `30a6583`, the last revision before that merge and the one every check here is green on: the fold is the same code on both sides of it.

Moving the pins to 527's current head needs tonk's own port of #524, which is separate from the peer port: ten effect definitions in `tonk-account` become an `Attenuation` plus an `Effect`, its three subscription commands need a `Subscription<method::Put>` namespace under the `put` verb in place of hand-written command strings, the access service's dispatch on those paths follows, and the worker and CLI have a handful of `archive().catalog()` and `RepositoryArchiveExt` sites behind that. A first workspace check stops at `tonk-account` with 23 errors, so the tail beyond it is unmeasured.

---

